PR #4, opened 2016-04-20
fix: end-of-frame detection.

### LINKED ISSUE #3: End-of-frame not properlly recognized
End-of-frame is currently assumed to be aligned with underlying stream termination. This might not always be true, e.g when some other information follow the compressed frame.

The following example

    use std::io::{Read, Write};

    let mut enc = encoder::Encoder::new(Vec::new(), 1).unwrap();
    enc.write_all(b"foo").unwrap();
    let mut compressed = enc.finish().unwrap();

    // Add footer/whatever to underlying storage.
    compressed.push(0);

    // Drain zstd stream until end-of-frame.
    let mut dec = decoder::Decoder::new(&compressed[..]).unwrap();
    let mut buf = Vec::new();
    dec.read_to_end(&mut buf).unwrap();
    assert_eq!(&buf, b"foo");

will return an instance of `Error { repr: Custom(Custom { kind: Other, error: StringError("Context should be init first") }) }`.

spaolacci:
Do not suppose end-of-frame is aligned with storage's end-of-file.
Handy when zstd frames are part of larger streams, used to return
some puzzling "Context should be init first" error.

Zstd natively recognizes frame termination and signal the situation
through the special `0` value of the returned variant:
"ZBUFF_decompressContinue returns either a hint to preferred nb
of bytes to use as input for next function call or 0 when a frame
is completely decoded, or an error code, which can be tested using
ZBUFF_isError()".

Closes #3.

### DIFF
--- a/src/stream/decoder.rs
+++ b/src/stream/decoder.rs
@@ -80,6 +80,10 @@ impl<R: Read> Decoder<R> {
 impl<R: Read> Read for Decoder<R> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
 
+        if self.offset > self.buffer.capacity() {
+            return Ok(0); // End-of-frame reached.
+        }
+
         let mut written = 0;
         while written != buf.len() {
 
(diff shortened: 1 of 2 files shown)


Gyscos:
So with multi-frame streams, does it stop after decoding the first frame?

spaolacci:
Yes it does stop correctly at the end of the current, wherever the frame is located w.r.t to its container stream.

Gyscos:
Perfect, thanks!

PR #8, opened 2016-08-16
Continue decompressing even if the underlying reader is at EOF

maximelenoir:
The only exit conditions for a call to read should be a full output buffer
or an end of zstd frame. Reaching EOF on the underlying reader is not a
valid exit condition as there might be some unflushed data in zstd
context, leading to an unexpected EOF. This patch does not prevent
calling the underlying reader several time at EOF.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,7 +9,7 @@
 //!
 //! # Example
 //!
-//! ```rust
+//! ```ignore
 //! extern crate zstd;
 //!
 //! use std::io;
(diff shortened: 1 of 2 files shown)


maximelenoir:
The build failed because of vec_map.

gyscos:
How, somehow I missed this PR - it would have saved me some time... ><

I noticed that too while updating zstd to 0.8.1, and ended up fixing it in f6550df14076afec64ee3a9c46166314467c0278 (I also added some checks to avoid looping endlessly if an early EOF is found), so I think this PR is no longer needed.

Thank you very much for your patch anyway!

PR #21, opened 2017-03-24
stream: add get_ref and get_mut to encoder and decoder types

sid0:
These turn out to be useful when inspecting or poking at the wrapped stream's internal state.


gyscos:
Looks good, thanks!

PR #23, opened 2017-04-08
Add try_finish

### LINKED ISSUE #22: Non-consuming version of finish() method?
Here's an interesting issue. I'm using zstd with [tokio-io](https://github.com/tokio-rs/tokio-io), which adds the notion of non-blocking writes signaled with `io::Error(io::ErrorKind::WouldBlock)`. Typically the bottom-most layer will need to be aware of the non-blocking nature of writes. Intermediate layers like `Encoder` can just pass the error message up and trust that higher layers will handle it.

However, with consuming methods like `finish()`, there's a wrinkle:

1. These methods cause writes to happen (a `flush()` before the `finish()` is insufficient because the `finish()` does a write of its own).
2. They will consume the encoder even if the error is a `WouldBlock` error (where the caller needs to retry the `finish()` after some time).

I'm really not sure how to handle this use case. A non-consuming version of `finish` would solve this issue but makes the API uglier :/

I'm also going to file an issue against tokio-io to talk about it there, because this seems like a general API deficiency.

sid0:
See the individual commit messages for the full details.

Closes #22.

### DIFF
--- a/src/stream/encoder.rs
+++ b/src/stream/encoder.rs
@@ -21,6 +21,16 @@ impl Drop for EncoderContext {
     }
 }
 
+#[derive(PartialEq)]
+enum EncoderState {
+    Accepting,
+    // Finished just makes sure writes are no longer accepted.
+    Finished,
+    StreamEnd,
+    // The "stream ended and flushed" state corresponds to the object being
+    // destroyed.
+}
+
 /// An encoder that compress and forward data to another writer.
 ///
 /// This allows to compress a stream of data
@@ -39,6 +49,7 @@ pub struct Encoder<W: Write> {
 
     // compression context
     context: EncoderContext,
+    state: EncoderState,
 }
 
 /// A wrapper around an `Encoder<W>` that finishes the stream on drop.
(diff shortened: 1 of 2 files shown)


src/stream/encoder.rs
```diff
@@ -183,34 +195,57 @@ impl<W: Write> Encoder<W> {
...
-        })?;
-        unsafe {
-            self.buffer.set_len(buffer.pos);
+    /// Attempts to finish the stream. You *need* to call this after writing your stuff.
+    ///
+    /// This returns the inner writer if the finish was successful, or the
+    /// object plus an error if it wasn't.
+    ///
+    /// `write` on this object will panic after `try_finish` has been called,
+    /// even if it fails.
```
gyscos: I wonder if it'd be a good idea to, instead of returning a half-broken `Self` in the error, return a `FinishingEncoder` or something that can *only* call `finish`?

sid0: I did consider that, but decided against it because users will have to deal with two different objects. So suddenly, rather than storing an `Encoder` in a struct, you'd have to store an enum with `Encoder(Encoder), Finishing(FinishingEncoder)`. It's not horrible by any means -- just a bit more complicated than this.

gyscos:
Looks good!

PR #24, opened 2017-04-10
encoder: avoid returning ErrorKind::Interrupted

sid0:
While `Interrupted` is theoretically a valid thing to return here, many
consumers deal poorly with it. It's much easier both for us and for consumers
to just loop until the internal buffer has been completely written out.

Other encoders like `bzip2-rs` and `flate2-rs` do the same thing.

### DIFF
--- a/src/stream/encoder.rs
+++ b/src/stream/encoder.rs
@@ -272,13 +272,8 @@ impl<W: Write> Write for Encoder<W> {
             panic!("write called after try_finish attempted");
         }
 
-        if self.offset < self.buffer.len() {
-            // If we still had some things to write, do it first.
-            self.offset += self.writer.write(&self.buffer[self.offset..])?;
-            // Maybe next time!
-            return Err(io::Error::new(io::ErrorKind::Interrupted,
-                                      "Internal buffer full"));
-        }
+        // Write any data pending in `self.buffer`.
+        self.write_from_offset()?;
 
         // If we get to here, `self.buffer` can safely be discarded.
 


gyscos:
From [the official doc on `Write.write`](https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.write): 
> A call to write represents at most one attempt to write to any wrapped object.

sid0:
Hmm, that's unfortunate. I guess the other crates are in violation of the contract then. :/

I'm going to discuss with a few people and get back.

sid0:
Thanks! I'll close this while this is resolved.

sid0:
I posted about this on the internals board: https://internals.rust-lang.org/t/std-io-write-a-call-to-write-represents-at-most-one-attempt-to-write-to-any-wrapped-object/5068

sid0:
Given the discussion on that thread, has your mind changed about this PR?

sid0:
(reopening to make sure this doesn't get lost =) )

gyscos:
What this does is say: "`write()` always tries to empty its buffer, ignoring Interrupt errors (but re-throwing any other error)." In this case, we should also use `write_from_offset` at the end of `write()`, and not just at the beginning of the next write.

My main concern is that this blurs the distinction between `write` and `write_all`.

sid0:
Yeah, that's true -- `write` will basically become a version of `write_all` that returns how many bytes were written.

sid0:
Hmm, what about this?

`write_from_offset` does `while self.offset < self.buffer.len()`.

What if `write` did the exact same thing that `write_from_offset` does, except use `if` instead of `while`.

gyscos:
I'm not sure I see the point ultimately.
* If `write` has some non-empty internal buffer, it *needs* to write it before writing any byte given to it. So if we want to avoid returning `Interrupted` or `Ok(0)`, we *need* to write the entire internal buffer in order to process at least one byte of input data.
* If we write the entire remaining internal buffer at the beginning of `write`, it doesn't make sense _not_ to do it at the end of write - we'd just be delaying the behavior by one step, without any real reason.

I don't think we should avoid returning an `Interrupted` error at all cost. I do understand that calling the inner `write` more than once is not completely forbidden, but I don't think there is a reason to do that here.

sid0:
That's a completely fair viewpoint.

Only other thing I wanted to point out was that this means that without either this PR or something like https://github.com/tokio-rs/tokio-io/pull/39, `zstd` is incompatible with `tokio`. Hopefully we'll get a resolution one way or the other.

gyscos:
This sounds like it could be done by a wrapper around `Encoder` that retries on interrupted errors?

sid0:
Yeah, that's what I'm doing internally in our use of `zstd-rs`, and that's basically what the PR is effectively doing. Still sucks for someone to have to learn this the hard way, though.

sid0:
Also, I was planning to implement `AsyncWrite` support behind an optional `tokio` feature as discussed earlier, and if we advertise compatibility with tokio we'd definitely want this to be resolved one way or the other.

gyscos:
Couldn't we put the re-try in the `AsyncWrite` implementation in this crate?

sid0:
`AsyncWrite` depends on `Write` (a bit of a wart IMO, and one that probably can't be fixed in Rust 1.x). This means that if `ZstdEncoder<W: AsyncWrite>` implemented `AsyncWrite` directly, we couldn't provide a different implementation for `write` AFAIK. The only way to make that work would be to have a separate `AsyncZstdEncoder<W: AsyncWrite>` with its own implementation of `write`.

sid0:
Yeah, I did a quick and dirty implementation in https://github.com/sid0/zstd-rs/commit/c1864f368f0a5ba1455b7b15494e876141cf367b and that fails as expected with

```
   Compiling zstd v0.4.7 (file:///data/users/sid0/rust/zstd-rs)
error[E0119]: conflicting implementations of trait `std::io::Write` for type `stream::encoder::Encoder<_>`:
   --> src/stream/encoder.rs:359:1
    |
326 | impl<W: Write> Write for Encoder<W> {
    | - first implementation here
...
359 | impl<W: AsyncWrite> Write for Encoder<W> {
    | ^ conflicting implementation for `stream::encoder::Encoder<_>`

error: aborting due to previous error

error: Could not compile `zstd`.
```

gyscos:
I see - I thought `AsyncWrite` would allow us to re-define its `write` method, but it directly uses the one from the `Write` implementation.

Well, if that's the only way, I'm ready to squash `Interrupted` errors here - I don't think people care that much about the difference between write and write_all.

gyscos:
I merged this PR, along with an extra commit to `write_from_offset` at the end of `write`. Hopefully tokio integration will be easier now!

gyscos:
... and it broke the "partial write" test, which expected the encoder to stop after the first byte written...

sid0:
Yeah I'll put up a fix in a bit.

sid0:
So :( `write_from_offset` at the end of `write` adds some complexity. Consider what happens if the first few attempts at writing succeed but then the write fails with a non-`Interrupted` error. In that case, at the moment we bubble up the error. This is wrong since writing out the input into the zstd compressor buffer did succeed.

Seems like we have two options:

1. Return `Ok(n)`, cache the last error, then return that error the next time `write` is called.
2. Don't do any writes to the inner writer at the end -- only do them the next time `write` is called, or if `flush` or `finish` are called.

(The old behavior where we would do one write at the end was also clearly wrong...)

gyscos:
Blast, I knew it was too easy <_<

> 1. Return Ok(n), cache the last error, then return that error the next time write is called.

This would be the natural solution, but I don't really like the complexity it adds.

> 2. Don't do any writes to the inner writer at the end -- only do them the next time write is called, or if flush or finish are called.

I like the simplicity of that. The main issue is that it doesn't feel as "natural" (always writing the data from the previous call).

I'd say we go with the second solution for now.

sid0:
OK. I did something slightly different in https://github.com/sid0/zstd-rs/tree/fix-write. It might be slightly more correct because it can also handle the case where for whatever reason the zstd compressor didn't accept any bytes.

gyscos:
I see. The encoder currently uses a output buffer sized using `ZSTD_CStreamOutSize`, which should guarantee that at least one block can always fit (and thus some input can always be consumed), but it might indeed be safer (and not very costly) to protect against this.

PR #25, opened 2017-04-12
encoder: write out buffer before starting flush or finish

sid0:
The problem here was that the buffer could still have some data in it at the
time it is overwritten. This would cause corruption, demonstrated by the
tests if run without the corresponding code changes.

None of the existing tests tested partial writes at all. To test this,
generalize `WriteWithReject` to also handle partial writes.


sid0:
FYI this is a critical corruption bug, so it would be good to have this landed and a new release made soonish.

sid0:
Thanks!

Issue #26, opened 2017-04-14
Add async integration


gyscos:
Work started in the [async](https://github.com/gyscos/zstd-rs/tree/async) branch.

TODO:

- [x] Add `AsyncWrite` to `stream::Encoder`
  - [x] Add tests
- [x] Add `AsyncRead` to `stream::Decoder`
  - [x] Add tests

sid0:
@gyscos would you like to do this yourself? If not I have someone else at Facebook interested in tackling this.

gyscos:
I would definitely appreciate help, especially for the tests.
I wrote the `AsyncWrite` implementation looking at how it was done in flate2, but I don't really know how to use it yet - I haven't had time to look at actual tokio code yet.

vladima:
I guess this issue now can be closed.

PR #28, opened 2017-04-24
proposal for AsyncRead/AsyncWrite implementation

vladima:
This PR is based on [async](https://github.com/gyscos/zstd-rs/issues/26) branch in this repo. 

- [x] - added implementation of `AsyncRead` to `Decoder`
- [x] - added tests for `AsyncRead` and `AsyncWrite`
- [x] - updated implementation of `read` to make sure that `ErrorKind::WouldBlock` is handled correctly

~~NOTE: we should probably add helpers to handle `ErrorKind::WouldBlock` in other methods besides `read` and `write` - i.e. in case of encoder `ErrorKind::WouldBlock` can be returned from `finish`.~~

// cc @sid0

### DIFF
--- a/src/stream/decoder.rs
+++ b/src/stream/decoder.rs
@@ -249,7 +249,7 @@ impl<R: Read> Decoder<R> {
 
 impl<R: Read> Read for Decoder<R> {
     fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        
+
         let mut in_buffer = zstd_sys::ZSTD_inBuffer {
             src: self.buffer.as_ptr() as *const c_void,
             size: self.buffer.len(),
--- a/src/stream/decoder.rs
+++ b/src/stream/decoder.rs
@@ -6,7 +6,6 @@ use zstd_sys;
 #[cfg(feature = "tokio")]
 use tokio_io::AsyncRead;
 
-
 struct DecoderContext {
     s: *mut zstd_sys::ZSTD_DStream,
 }
@@ -24,7 +23,6 @@ impl Drop for DecoderContext {
     }
 }
 
-
 /// A decoder that decompress input data from another `Read`.
 ///
 /// This allows to read a stream of compressed data
@@ -41,6 +39,9 @@ pub struct Decoder<R: Read> {
 
     // `true` if we should stop after the first frame.
     single_frame: bool,
+
+    // 'false' if end of frame was not yet reached
+    end_of_frame: bool,
 }
 
 impl<R: Read> Decoder<R> {
@@ -77,6 +78,7 @@ impl<R: Read> Decoder<R> {
             offset: 0,
             context: context,
             single_frame: false,
+            end_of_frame: false,
         };
 
         Ok(decoder)
@@ -84,6 +86,7 @@ impl<R: Read> Decoder<R> {
 
     fn reinit(&mut self) -> io::Result<()> {
         parse_code(unsafe { zstd_sys::ZSTD_resetDStream(self.context.s) })?;
+        self.end_of_frame = false;
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


gyscos:
If I understood correctly, we don't want to retry on WouldBlock, but allow to recover safely from it. This is what `try_finish` does for the `Encoder; does another method needs something like this?

src/stream/decoder.rs
```diff
@@ -20,6 +23,18 @@ impl Drop for DecoderContext {
     }
 }
 
+#[derive(PartialEq, Copy, Clone)]
+enum RefillBufferHint {
```
gyscos: Could you add a few coments to explain what those hints are for? Thanks!

src/stream/decoder.rs
```diff
@@ -125,14 +152,109 @@ impl<R: Read> Decoder<R> {
...
+impl<R: Read> Decoder<R> {
+    /// This function handles buffer_refill state of the read operation
+    /// It returns true if read operation should be stopped and false otherwise
+    fn handle_refill(&mut self, hint: RefillBufferHint, 
+        in_buffer: &mut zstd_sys::ZSTD_inBuffer, out_buffer: &mut zstd_sys::ZSTD_outBuffer)-> Result<bool, io::Error> {
+
+        let refilled = match self.refill_buffer(in_buffer) {
+            Err(ref err) if out_buffer.pos > 0 && err.kind() == io::ErrorKind::WouldBlock => {
+                // underlying reader was blocked but we've already put some data into the output buffer
+                // we need to stop this read operation so data won' be lost
```
gyscos: What happens if another Error, like Interrupted, is returned?

vladima: it will interrupt the read similar to current behavior. As an option we can have a loop and retry similar to what `write_from_offset` is doing today.

gyscos: Indeed, the current behaviour is incorrect.

vladima:
Please disregard the note, I've not understood the problem well enough.

As another question: currently in tests there is an [implementation](https://github.com/gyscos/zstd-rs/pull/28/files#diff-f30781bdb71d584215e104444e2c92dbR461) of the future that will handle `Encoder::finish` operation and it feels that it might be the code that user will also have to write if they want to use `Encoder` with futures. Should it be pulled out of tests and conditioned on the `tokio` feature?

src/stream/decoder.rs
```diff
@@ -121,16 +148,120 @@ impl<R: Read> Decoder<R> {
...
+                return Ok(false);
+            }
+            if is_end_of_frame && self.single_frame {
+                // at the end of frame and we know that this frame is the only one
+                // stop
+                self.state = DecoderState::Completed;
+                return Ok(true);
+            }
+        }
+        return Ok(true);
```
sid0: Just `Ok(true)`

src/stream/decoder.rs
```diff
@@ -121,16 +148,120 @@ impl<R: Read> Decoder<R> {
...
+                }
+                else {
+                    // no more data - we are done
+                    self.state = DecoderState::Completed;
+                    return Ok(true);
+                }
+            }
+        }
+
+        return Ok(false);
```
sid0: Just `Ok(false)` -- `return` at the end isn't required and is not considered appropriate

src/stream/decoder.rs
```diff
@@ -142,54 +273,139 @@ impl<R: Read> Read for Decoder<R> {
...
+    }
+}
 
-            if res > 1 && input_exhausted {
-                // zstd keeps asking for more, but we're short on data!
-                return Err(io::Error::new(io::ErrorKind::UnexpectedEof,
-                                          "incomplete frame"));
+#[cfg(feature = "tokio")]
+impl<R: AsyncRead> AsyncRead for Decoder<R> {
+    unsafe fn prepare_uninitialized_buffer(&self, _buf: &mut [u8]) -> bool {
```
sid0: I think you can just skip implementing this?

gyscos: Since we never read the output buffer, I suppose it's safe to avoid zeroing it out, improving performance a little bit.

vladima: that is exactly my reason

vladima:
@gyscos do you have any other questions or comments?

gyscos:
PR was merged (not sure why github isn't detecting it).
The dedicated `Finish` future is a good point - it could be returned by `Encoder.finish_future(self)` or something like that...

sid0:
Thanks!

BTW I'm currently working on some randomized (QuickCheck) test support for partial reads/writes, `WouldBlock` and `Interrupted`. Hopefully I'll get it out as an open source crate in a couple of weeks so you can just start using that instead :)

PR #29, opened 2017-05-30
use partial-io for I/O tests

sid0:
We recently released the `partial-io` crate, which adds convenience wrappers
for partial reads and writes.  It's basically a more formal version of a couple
of helpers we already had.

`partial-io` also supports `quickcheck` for random test cases. That could
eventually be used as well, as in e.g.
https://github.com/alexcrichton/bzip2-rs/pull/20.

What do you think?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ futures = { version = "0.1", optional = true }
 
 [dev-dependencies]
 clap = "2.6.0"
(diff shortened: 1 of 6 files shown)


gyscos:
I suppose some `cargo fmt` got mixed in there, but it looks good otherwise!
Thank you for the work!

sid0:
(I made the formatting changes in a separate commit)

PR #30, opened 2017-06-05
stream: switch to randomized testing with partial-io

sid0:
This is pretty straightforward and a good demonstration that zstd
doesn't break down when confronted with transient errors.


gyscos:
Looking good, thanks!

PR #31, opened 2017-06-05
travis: add CI for the optional tokio feature

sid0:
This makes sure the project builds both with tokio and without.


gyscos:
Thanks!

PR #32, opened 2017-06-23
Make zstd contexts and streams support Send, and dicts support Send + Sync

sid0:
According to @Cyan4973, streams and contexts support `Send`, and dicts
support both `Send` and `Sync`.

Also include compile-time checks to make sure the higher-level stream and block APIs support `Send`.

### DIFF
--- a/src/block/compressor.rs
+++ b/src/block/compressor.rs
@@ -62,3 +62,9 @@ impl Compressor {
         Ok(buffer)
     }
 }
+
+fn _assert_traits() {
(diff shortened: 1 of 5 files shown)


gyscos:
Thanks!

gyscos:
Do you need a release with this, or can it wait until the release for zstd-1.3?

sid0:
A new release would be great!

gyscos:
Published [version 0.4.11](https://crates.io/crates/zstd).

PR #35, opened 2017-11-12
Update `bindgen` dependency to 0.31.X

fitzgen:
The `hide_type` method has been renamed to `blacklist_type` to match the CLI flag.

The bindings seem to work just fine with `generate_comments` for me now.


gyscos:
Thanks! I thought I did something like that, but apparently not...

PR #36, opened 2017-11-21
adjust build

BusyJay:
So we can call `cmake::Build().register_dep("ZSTD")`.

### DIFF
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -4,6 +4,9 @@ extern crate bindgen;
 extern crate gcc;
 extern crate glob;
 
+use std::{env, fs};
+use std::path::PathBuf;
+
 
 #[cfg(feature = "bindgen")]
 fn generate_bindings() {
@@ -78,6 +81,13 @@ fn compile_zstd() {
 
     // Compile!
     config.compile("libzstd.a");
+
+    let src = env::current_dir().unwrap().join("zstd").join("lib");
+    let dst = PathBuf::from(env::var_os("OUT_DIR").unwrap());
+    let include = dst.join("include");
+    fs::create_dir_all(&include).unwrap();
+    fs::copy(src.join("zstd.h"), include.join("zstd.h")).unwrap();
+    println!("cargo:root={}", dst.display());
 }
 
 fn main() {
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -87,6 +87,9 @@ fn compile_zstd() {
     let include = dst.join("include");
     fs::create_dir_all(&include).unwrap();
     fs::copy(src.join("zstd.h"), include.join("zstd.h")).unwrap();
+    let dict_builder = include.join("dictBuilder");
+    fs::create_dir_all(&dict_builder).unwrap();
+    fs::copy(src.join("dictBuilder").join("zdict.h"), dict_builder.join("zdict.h")).unwrap();
     println!("cargo:root={}", dst.display());
 }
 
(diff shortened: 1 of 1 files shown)


gyscos:
Hi, and thanks for the PR!

I'm not very familiar with using Cmake with Rust, so could you explain a bit what this enables?
For instance, this library uses functions declared in other headers, like `zdict.h` - should those be copied as well?

BusyJay:
This pr sets a metadata root to building directory, which will [be set](http://doc.crates.io/build-script.html#the-links-manifest-key) as environment variable `DEP_ZSTD_ROOT`. When calling `register_dep`, the directory will [be added](http://alexcrichton.com/cmake-rs/cmake/struct.Config.html#method.register_dep) to `CMAKE_PREFIX_PATH`, which is used to search headers and libraries.

Only public interface is needed, so I just copy `zstd.h`.

gyscos:
I'm still curious what use-case would benefit from that. Is this when integrating this rust library in a larger, multiple-languages project?

As a note, `zstd.h` is not the only public interface: `zdict.h` declares dictionary-training methods not present in `zstd.h`, used by the `dict` module for instance.

BusyJay:
For example, rocksdb (a kv engine) optionally depend on zstd. To enable zstd support for its rust wrapper, it can either just add zstd-sys as a dependency or add zstd c sources as a submodule then write the build.rs (again).

BusyJay:
Another example, [git-sys](https://github.com/alexcrichton/git2-rs/tree/master/libgit2-sys) depends on ssh2 and libz, but it doesn't have to add each as a submodule, but just depend on their *-sys packages instead. It's actually a [study case](http://doc.crates.io/build-script.html#building-libgit2) listed in crates.io.

gyscos:
I'm beginning to understand: rust-rocksdb wants to build rocksdb using cmake, and have it auto-detect the zstd library?
So you don't need the rust part at all, you just want to have a zstd library that cmake can find?

In that case, `zdict.h` is definitely useful: in [`compression.h`](https://github.com/facebook/rocksdb/blob/master/util/compression.h#L39), rocksdb explicitely includes it.

Not sure if that's relevant, but zstd itself [includes a CMakeList file](https://github.com/facebook/zstd/blob/master/build/cmake/CMakeLists.txt).

BusyJay:
Thanks for the information. I updated the pr, `zdict.h` is copied now. The `CMakeList` is useless unless there is a submodule. Although this crate may take advantage of cmake and skip the compilation details.

gyscos:
Well, thanks for the explanations!
I'm wondering if this is enough, though. I suppose the goal is to make a structure recognizable by cmake, I think the `libzstd.a` library itself needs to be copied to `$OUT_DIR/lib`?

Looking at the current `libssh2-sys`, it seems to just be using `cmake`, and doesn't copy these files. [A previous version](https://github.com/alexcrichton/ssh2-rs/blob/5688e2bd912bb1ad8de261e5849f58b4857a6818/libssh2-sys/build.rs) apparently was:
* On unix, just using `make install` commands to copy the library and the include files in the `OUT_DIR`, thus ensuring a structure the cmake would like.
* On windows, manually copy the library and header files to `$OUT_DIR/lib` and `$OUT_DIR/include` respectively.

Also, the `zdict.h` must go directly in the `include` directory, not in a subfolder.

BusyJay:
> ... I think the libzstd.a library itself needs to be copied to $OUT_DIR/lib?

Not necessary, as long as the library exist in CMAKE_PREFIX_PATH, which is set by `cargo:root`.

> ..  it seems to just be using cmake, and doesn't copy these files. ...

cmake-rs will set the `cargo:root` automatically. Hence library can be loaded automatically.

> .. the zdict.h must go directly in the include directory, not in a subfolder.

Fixed.

This is tested in pingcap/rust-rocksdb#159. But we only use rocksdb 5.7.3 now, which doesn't include `zdict.h` yet.

gyscos:
Ah, I see. Looks good then! Thanks again!

PR #38, opened 2018-01-15
typo in pub method name (create-ddict)

FauxFaux:
Unfortunately it's a public method, so should probably be a semver bump to release.


gyscos:
Ouch, thanks a lot for noticing!

PR #39, opened 2018-01-15
typo in pub method name (csict -> cdict)

FauxFaux:


gyscos:
Whoops, thank you very much!

Issue #37, opened 2018-01-14
Support for --no-dictID


FauxFaux:
Currently, I believe, there's no way to:
 * specify the dictionary ID for `zstd::Encoder::with_dictionary`.
 * omit the dictionary ID from the stream, as the `zstd --no-dictID` command-line option does.

I'm after the second; as currently I get an extra null byte in my output when using dictionaries; the messages are actually bigger than without a dictionary.

Please add them!

gyscos:
Thanks for the input!

It seems that:
* The `dictID` can be specified with `ZSTD_cParameter::ZSTD_p_dictIDFlag`.
* `noDictIDFlag` is a parameter of `ZSTD_frameParameters`, which can be given:
    * in `ZSTD_compress_using_CDict_advanced` or `ZSTD_initCStream_using_CDict_advanced`
    * as part of `ZSTD_parameters`, in `ZSTD_compressed_advanced` or `ZSTD_initCStream_advanced`

The `zstd` crate itself relies on `zstd-safe`, so that's where these feature should first appear.

Regarding the first point, I think this is what #40 helps with?
Regarding the second point, it's part of the "advanced" APIs that I haven't ported to `zstd-safe` yet.

FauxFaux:
I believe, but haven't tested, that you can `setParamters` `ZSTD_p_dictIDFlag` on an existing `cctx` and it will work. Hence working on `setParameters`.

Maybe some of the sensible options could be exposed as builder-like methods on context creation?

I didn't see a way to fiddle with the `cctx` inside a `cstream`, so it can't fit perfectly into the current API, unless I missed something.

phiresky:
Another thing that would be useful is allowing creation and decoding of magic-less headers (https://github.com/facebook/zstd/issues/591 https://github.com/facebook/zstd/pull/1080 ). I'm not sure if this is done with the same API as the dictid setting would be so I'm not sure if it's worth having as a separate issue.

phiresky:
I've opened a PR here #98 to fix both dictid as well as magic header removal.

PR #40, opened 2018-01-15
Partial safe wrapper for ZSTD_CCtx_setParameter

FauxFaux:
I started writing a safe wrapper for `ZSTD_CCtx_setParameter`, which would allow users of the lower level API to work around #37. I've suddenly realised it probably doesn't actually help me for my code, though, so I stopped. Maybe I'll start again when I realise I've made a mistake.

Compiles, doesn't break any tests, but isn't run by any tests.

Hopefully it will be useful for someone.

### DIFF
--- a/zstd-safe/src/lib.rs
+++ b/zstd-safe/src/lib.rs
@@ -807,6 +807,159 @@ pub fn reset_dstream(zds: &mut DStream) -> usize {
     unsafe { zstd_sys::ZSTD_resetDStream(zds.0) }
 }
 
+pub type Unsigned = libc::c_uint;
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum FrameFormat {
+    /// zstd frame format, specified in zstd_compression_format.md (default)
+    One,
+
+    /// Variant of zstd frame format, without initial 4-bytes magic number.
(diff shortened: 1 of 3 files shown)


FauxFaux:
`bindgen` has changed behaviour since the last time the bindings.rs was generated, and for good reason.

I've elected to tell `bindgen` to use the old behaviour (and stick with my current code), as this maintains the most compatibility. Should definitely go to the new `bindgen` way at some point, though.

gyscos:
Thank you very much for this work!

Regarding the bindgen change, I see `enum` are now translated into a bunch of `const` values instead (with weird prefixed names). We'll see later about that, but for now rustified enums are not a problem.

JakubOnderka:
@gyscos Is there any plan to merge that?

gyscos:
Sure, I was mostly waiting to see if anyone needs it.

FauxFaux:
I've updated the commits to do what the newer `bindgen` you're using does. I don't really know or understand the differences (I lost interest about six months ago!). It compiles, and the tests pass, both with and without `--features bindgen`.

It seems like it should be possible to drop that `rustify_enums` line now, but if you take it out, something goes wrong with `--features`, which doesn't seem worth fixing.

gyscos:
Thanks!

PR #44, opened 2018-06-28
Do not build ZSTD deprecated features

JakubOnderka:
Deprecated methods are not used in zstd-safe, so it is not necessary to build it


gyscos:
Great idea, I overlooked that!
Thanks for the PR!

PR #45, opened 2018-06-29
Support for prepared dictionaries for stream Encoder and Decoder

### LINKED ISSUE #17: Use `ZSTD_createCDict` to re-use dictionary in (De)Compressor
Better performance when encoding multiple blocks with the same dict.

JakubOnderka:
Prepared dictionaries are much faster when processing multiple streams with same dictionary, especially for short streams with low compression level.

Stream of 196 bytes compressed with level 1:
```
test tests::bench_with_dictionary              ... bench:      21,634 ns/iter (+/- 9,484)
test tests::bench_with_prepared_dictionary     ... bench:       4,703 ns/iter (+/- 3,200)
```

Fixes #17

### DIFF
--- a/src/dict.rs
+++ b/src/dict.rs
@@ -21,6 +21,46 @@ use std::io::{self, Read};
 use std::path;
 use zstd_safe;
 
+/// Prepared dictionary for compression
+pub struct EncoderDictionary<'a> {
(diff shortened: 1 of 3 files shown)


gyscos:
Thanks for the work!

PR #47, opened 2018-09-29
Removed dependency to libc from base package

JakubOnderka:
libc::size_t is replaced by usize


src/lib.rs
```diff
@@ -22,9 +22,6 @@
 //! ```
 //!
 //! [zstd]: https://github.com/facebook/zstd
-#![deny(missing_docs)]
```
gyscos: Let's not remove that one :)

gyscos:
I suppose `libc::size_t` is guaranteed to always be `usize` on all platform?

I may later update `zstd-safe` to return `usize`, so the `zstd` crate itself doesn't have to convert from `size_t` to `usize`.

src/lib.rs
JakubOnderka: Sorry, I thought it belongs to libc crate 😄 I will fix it

JakubOnderka:
> I suppose libc::size_t is guaranteed to always be usize on all platform?

I think it is the same. But also all functions from `zstd-safe` crate returns `usize` and input for `parse_code` functions are just outputs from `zstd-safe`. So it doesn't make sense to use `libc::size_t` just at this place.

src/lib.rs
JakubOnderka: Fixed

PR #50, opened 2018-10-26
make it possible to build `zstd` without `zstd-safe/std`

nickelc:
A [build][hpk-build] started to fail with zstd `0.4.21+zstd.1.3.7` on `i686-pc-windows-gnu` and i had to pin zstd to the previous version because of a bindgen issue and the forced `zstd-safe/std` feature.
Everything works fine for `i686-pc-windows-*` without the bindgen feature.

I haven't tried windows on travis yet. so, here the [build jobs][zstd-build] for zstd on appveyor.
- *-pc-windows-gnu builds without std 
- x86_64-pc-windows-gnu builds with std
- i686-pc-windows-gnu fails with std

[hpk-build]: https://ci.appveyor.com/project/nickelc/hpk/builds/19811691/job/ogt35xy4gn0qkxm4
[zstd-build]: https://ci.appveyor.com/project/nickelc/zstd-rs/builds/19827928

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ readme = "Readme.md"
 travis-ci = { repository = "gyscos/zstd-rs" }
 
 [dependencies]
-zstd-safe = { path="zstd-safe", version = "1.4.5", default-features = false, features=["std"] }
+zstd-safe = { path="zstd-safe", version = "1.4.5", default-features = false }
 tokio-io = { version = "0.1", optional = true }
 futures = { version = "0.1", optional = true }
 
@@ -27,7 +27,8 @@ humansize = "1.0"
 walkdir = "2.2.5"
 
 [features]
-default = ["legacy"]
+default = ["legacy", "std"]
 legacy = ["zstd-safe/legacy"]
 bindgen = ["zstd-safe/bindgen"]
+std = ["zstd-safe/std"]
 tokio = ["tokio-io", "futures", "partial-io/quickcheck", "partial-io/tokio"]


gyscos:
Interesting, the build only fails for `zstd-sys` when the `bindgen` feature is used?

The `zstd` crate itself doesn't care about `zstd-safe` (or `zstd-sys`) usage of `std` vs `libc`, so it probably shouldn't ask for `zstd-safe/std`. This feature is only used for webassembly, where `libc` is empty but `std` is present.

I hoped that both wasm and windows could build with the default features, but it seems impossible for now. So until we fix `bindgen` on windows, we default back to not asking for `zstd-safe/std` at all, and let wasm users require it.

gyscos:
Could it be related to https://github.com/rust-lang-nursery/rust-bindgen/issues/1276#issuecomment-375848230 ?

nickelc:
yeah, i saw this issue. 
The installed LLVM version in `PATH` is [7.0.0 x64](https://www.appveyor.com/docs/windows-images-software/#llvm)

gyscos:
It seems to say `C:\Program Files\LLVM\bin\libclang.dll` is a 64-bits DLL while it is supposed to be a i686 build.
Is there a 32-bits DLL somewhere else? Either way, it sounds like a [clang-sys](https://github.com/KyleMayes/clang-sys) issue.

nickelc:
I'm not familiar with cross compiling and the LLVM/clang stack.
So, it basically means to install llvm manually for the specific target?
```cmd
curl -sSf http://releases.llvm.org/%LLVM_VERSION%/LLVM-%LLVM_VERSION%-%LLVM_TARGET%.exe -o LLVM.exe
7z x LLVM.exe -oC:\LLVM
set LIBCLANG_PATH=C:\LLVM\bin
```

nickelc:
ok, build works with installed win32 version
https://ci.appveyor.com/project/nickelc/zstd-rs/builds/19836343

gyscos:
:tada:

PR #52, opened 2018-12-05
Make zstd_safe really safe

JakubOnderka:
One part of being safe is to check error codes. But current implementation always return usize and return code must be checked in zstd-rs crate. This commit changes the behaviour that zstd-safe functions returns SafeResult.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,18 +49,10 @@ pub use zstd_safe::CLEVEL_DEFAULT as DEFAULT_COMPRESSION_LEVEL;
 #[doc(no_inline)]
 pub use stream::{decode_all, encode_all, Decoder, Encoder};
 
-/// Parse the result code
-///
-/// Returns the number of bytes written if the code represents success,
-/// or the error message otherwise.
-fn parse_code(code: usize) -> io::Result<usize> {
-    if zstd_safe::is_error(code) == 0 {
-        Ok(code)
-    } else {
-        let msg = zstd_safe::get_error_name(code);
-        let error = io::Error::new(io::ErrorKind::Other, msg.to_string());
-        Err(error)
-    }
+/// Returns the error message as io::Error based on error_code.
+fn map_error_code(code: usize) -> io::Error {
+    let msg = zstd_safe::get_error_name(code);
+    io::Error::new(io::ErrorKind::Other, msg.to_string())
 }
 
 // Some helper functions to write full-cycle tests.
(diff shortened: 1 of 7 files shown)


zstd-safe/src/lib.rs
```diff
@@ -81,6 +81,20 @@ pub const FRAMEHEADERSIZE_MIN: u32 = zstd_sys::ZSTD_FRAMEHEADERSIZE_MIN;
...
 pub const JOBSIZE_MIN: u32 = zstd_sys::ZSTDMT_JOBSIZE_MIN;
 
+type SafeResult = Result<usize, u32>;
+
+/// Parse the result code
+///
+/// Returns the number of bytes written if the code represents success,
+/// or the error message code otherwise.
+fn parse_code(code: usize) -> SafeResult {
+    let error_code = unsafe { zstd_sys::ZSTD_isError(code) as u32 };
```
gyscos: I think `ZSTD_isError` just returns a "boolean" (a 1 or 0 integer), and not the code itself. C doesn't have booleans, which is why `ZSTD_isError` returns a "unsigned integer".
You probably should use `zstd_safe::is_error` which returns a `bool` and is less confusing.

```rust
type SafeResult = Result<usize, usize>; // Why convert to `u32`?

fn parse_code(code: usize) -> SafeResult {
    if is_error(code) == 0 {
        Ok(code)
    } else {
        Err(code)
    }
} 
```

src/lib.rs
```diff
@@ -49,18 +49,10 @@ pub use zstd_safe::CLEVEL_DEFAULT as DEFAULT_COMPRESSION_LEVEL;
...
-fn parse_code(code: usize) -> io::Result<usize> {
-    if zstd_safe::is_error(code) == 0 {
-        Ok(code)
-    } else {
-        let msg = zstd_safe::get_error_name(code);
-        let error = io::Error::new(io::ErrorKind::Other, msg.to_string());
-        Err(error)
-    }
+/// Returns the error message as io::Error based on error_code.
+fn map_error_code(code: u32) -> io::Error {
```
gyscos: Actually most (all?) error codes won't fit in `u32` since they are defined as negative `size_t` values (so they'll be very close to usize::max_value`.

gyscos:
Sure, I guess we can make this check in `zstd-safe` (the "safe" was for memory safety, which is not violated by returning raw codes). Just a few comments regarding the result type and the parse function.

zstd-safe/src/lib.rs
JakubOnderka: True, I pushed more fixed and more clear solution.

src/lib.rs
JakubOnderka: `u32` was mistaken, because I thought `ZSTD_isError` returns error code. Now it is fixed.

gyscos:
Looks cool, thanks!

Issue #58, opened 2019-01-09
Duplicate symbols error


terrelln:
I'm not a rust expert, so forgive my terminology. This crate should hide all libzstd symbols in the `rlib`. Otherwise, when using the `zstd-sys` module, and linking with other C/C++ code that also uses zstd, you will get duplicate symbol warnings.
    
You can get zstd to hide all of its symbols with these compiler flags:

```
-DZSTDLIB_VISIBILITY=
-DZDICTLIB_VISIBILITY=
-DZSTDERRORLIB_VISIBILITY=
-fvisibility=hidden
```

See the example in [python-zstandard](https://github.com/indygreg/python-zstandard/blob/master/setup_zstd.py#L157).

gyscos:
Good point! I'll add these flags.

gyscos:
I published `zstd-sys` 1.4.8, which should add these options.
Does it solve the problem?
Even better, do you have a test case for this we could add/document somewhere?

terrelln:
You can `nm -g zstd-sys.rlib | grep -e ZSTD -e ZDICT -e HUF -e FSE` (or whatever the rlib is called). If there are no symbols listed as `T` then the patch likely worked. 

Sorry, I spent some more time on this, but forgot to update this issue. The visibility setting (`-fvisibility=hidden`) sadly doesn't work because afaik the `rlib` is a static object, not a shared library.

I'm not sure how to fix this issue. I'm looking for help from Rust users internally. If we find a fix, I'll update you here.

gyscos:
Than you for the analysis!

When you say `zstd-rs` uses the unstable API, which part are you talking about? Do you mean the experimental bits in `zstd-sys` and `zstd-safe`? The top-level `zstd` crate itself ideally shouldn't be using these.

From what I understand, we can link zstd either statically or dynamically; and we can either use the local submodule, or an externally-provided one (maybe the system version). This results in 4 possible situations, though the most common are probably statically-linked locally-built version, and dynamically-linked system-provided version. Dynamically-linking the locally-built library doesn't sound very interesting, but I suppose statically-linked provided-version might be desirable (as you mention, when using pkgconfig)?

I understand that dynamic linking precludes us from using/exposing unstable part of zstd (like some of the experimental parameters). When statically linking a provided version of zstd, I'm not so sure - we might need to re-run bindgen with the provided header, but I think allowing unstable features would be safe in this case.

terrelln:
[...]

PR #59, opened 2019-01-18
Add pkg-config feature to use pre-installed libzstd

jsgf:
This will use pkg-config to locate the existing library. Use in
combination with bindgen to generate a binding from that header.

The `non-cargo` feature will suppress emitting metadata for cargo
(for cases where cargo isn't doing the final link).

Addresses issue https://github.com/gyscos/zstd-rs/issues/58


gyscos:
Thanks a lot for the PR!
I'll have a look shortly, but it seems simple enough. :D

PR #60, opened 2019-01-23
Remove unneeded blob crate

messense:


gyscos:
Sorry for the delay, and thanks for the PR!

PR #68, opened 2019-10-21
Copy public zstd_errors.h into output directory

benesch:
The common/zstd_errors.h header file is also part of the public
interface of the zstd library. Copy it into the output directory, like
the upstream build system does [0], so that downstream dependencies can
find it.

[0]: https://github.com/facebook/zstd/blob/0f9866add2df8e824c96f67fc49a7d5b69ad3317/lib/Makefile#L275


gyscos:
Thanks for the PR!

benesch:
Thanks for the quick merge! Any idea when the next release will be
published?

PR #71, opened 2019-11-22
Export includes paths as cargo metadata

bkchr:
Depending projects will be able to use this by looking into the env
varilabe `DEP_ZSTD_INCLUDE`. This variable will contain a `;` seperated
list of include paths.

### DIFF
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -6,9 +6,12 @@ extern crate pkg_config;
 
 extern crate cc;
 extern crate glob;
+extern crate itertools;
 
 use std::path::PathBuf;
 use std::{env, fs};
+
+use itertools::Itertools;
 
 #[cfg(feature = "bindgen")]
 fn generate_bindings(defs: Vec<&str>, headerpaths: Vec<PathBuf>) {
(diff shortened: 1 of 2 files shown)


zstd-safe/zstd-sys/build.rs
```diff
@@ -133,21 +136,23 @@ fn compile_zstd() {
...
+            env::var("CARGO_MANIFEST_DIR")
+                .expect("Manifest dir is always set by cargo"),
+        );
+
         compile_zstd();
-        (vec![], vec![PathBuf::from("zstd/lib")])
+        (vec![], vec![manifest_dir.join("zstd/lib")])
     };
 
+    println!("cargo:include={}", headerpaths.iter().map(|p| p.display().to_string()).join(";"));
```
gyscos: Do we need `Display` for that? Why not just `iter().filter_map(PathBuf::to_str)` or something like that?

bkchr: But doesn't that mean we could loose some parts? The documentation says that the display mitigates non Unicode paths.

gyscos: True, but if a path contains non-unicode parts, will this actually work as an include path? I would think the "unicode-safety" here means a lossy conversion that may work for pretty-printing, but might not work as an include path.

PR #75, opened 2020-01-01
Misuse of BufRead might lead to hangs

FauxFaux:
(This is a bug report, with attached code.. making it a PR? Eh, thanks github.)

The `Reader` utility apparently uses `BufRead`'s `fill_buf` incorrectly. It repeatedly calls `fill_buf` until the codec makes progress. However, typical `BufRead` implementations, including `std::io::BufReader` will *never* read more data from the underlying reader, unless the client has consumed every single byte.

> Returns the contents of the internal buffer, filling it with more data from the inner reader if it is empty.

The test tries to demonstrate by this by having an `Operation` which wants two bytes, but a `Read` which only returns one byte at a time. It hangs indefinitely. I didn't really check the test is correct, but it demonstrates the problem I am thinking about.

I've not actually hit this bug with `zstd-rs`, but have written the same bug in other code. Is `zstd` itself doing internal buffering, helping us not hit this (and making the retry loop pointless)? I guess the test could be extended to check that.

### DIFF
--- a/src/stream/zio/reader.rs
+++ b/src/stream/zio/reader.rs
@@ -159,8 +159,9 @@ where
 
 #[cfg(test)]
 mod tests {
-    use super::Reader;
-    use std::io::{Cursor, Read};
+    use super::{Operation, Reader};
+    use std::io::{self, Cursor, Error, Read};
+    use zstd_safe::{InBuffer, OutBuffer};
 
     #[test]
     fn test_noop() {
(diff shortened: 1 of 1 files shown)


gyscos:
Hi, and thanks for the report!

Indeed, operations are currently expected to always either read at least one byte, or write at least one byte per step. This is indeed (I think) what is guaranteed by the zstd API we're using.

A more general implementation that would be robust to different operations might be interesting, especially if we want to split it in a separate crate, but might be trickier to design, especially if we want to avoid unnecessary copies for performance reasons.

gyscos:
TLDR: yes, the zstd library has its own internal buffer, and will always consume some input.

Issue #64, opened 2019-08-08
Parallel (de)compression


thedrow:
Most bindings I know support parallel (de)compression of Zstd blocks.
It'd be nice to have that API in these bindings as well.

gyscos:
Agreed! I think moving to the advanced API will help enable that.

gyscos:
I updated `zstd-safe` to better expose the advanced options, and use it in the top-level crate for 2 options so far: checksum, and parallel compression.
Note that I'll need to enable multi-threading when building `zstd` in `zstd-sys`...

JJJollyjim:
Does that mean this should be working in the current version?

I have `features = ["zstdmt"]` on `zstd-safe` in my `Cargo.toml`, but I can't get past "Unsupported parameter" when calling `encoder.multithread`.

andoriyu:
@JJJollyjim have you ever solve this?

andoriyu:
@gyscos I've figured it out. PR incoming.

PR #84, opened 2020-04-01
fix: Set ZSTD_MULTITHREAD when built with zstdmt

andoriyu:
Advanced features like multithreading are only available when libzstd
is statically linked per documentation. pkg-config is not very good
for static linking. That means users have to rely on this crate to build it
correctly.

Current `build.rs` is not passing `ZSTD_MULTITHREAD` flag to cc and
because of that resulting library built without multithreading support.

Resolves: #64


gyscos:
Thanks for the work! Will try to run some benchmarks to see the nice gains

PR #90, opened 2020-06-15
Generate experimental dictionary bindings

indygreg:
dictBuilder/zdict.h contains additional experimental zstd APIs.
This includes low-level APIs for training dictionaries with full
control over parameters.

I need access to these APIs in order to implement a Rust backend
for my Python bindings to zstandard project
(https://github.com/indygreg/python-zstandard).

For reasons unknown to me, the dictionary experimenting APIs are
gated behind a separate preprocessor define. Simply adding this
define is enough to cause the bindings to be generated.


gyscos:
Hi, and thanks for the work!
Everything looks good, thanks again!

PR #92, opened 2020-07-28
using of target_arch/_os to detect, if the platform has a libc

stlankes:
Beside wasm32 also RustHermit doen't have a classical C library.
Using of target_arch/_os to detect these platforms. It simplifies
the usage of this crate.


gyscos:
Hi, and thanks for the PR!

I think I'd rather keep `std` as a feature, and turn it on automatically in a `build.rs` depending on the target (and do the same in `zstd-sys`).

stlankes:
Yes, I see the benefits of using the feature `std`. Do you think more of such a solution?

gyscos:
Looking good - thanks again for the work!

PR #97, opened 2020-11-28
Update bindgen to 0.56

decathorpe:
This bindgen update does not require other code changes, as far as I can tell. A local build of zstd-sys with bindgen 0.56 was successful.


gyscos:
Thanks for the PR!

PR #98, opened 2020-12-19
add more config functions for better compression for small data

### LINKED ISSUE #37: Support for --no-dictID

phiresky:
adds configurability of the following:

* dictid inclusion (saves 4 bytes when using dicts)
* include_magicbytes (saves 4 bytes always)
* include_contentsize (unsure)

Fixes #37 .


gyscos:
Looks great, thanks for the work!

PR #103, opened 2021-01-07
zstd-sys: avoid dependencies on itertools and glob

benesch:
@gyscos not sure if you're open to this, but figured I'd whip it up. I'm on a quest downstream to lighten up some dependencies.

### DIFF
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -5,8 +5,8 @@ extern crate bindgen;
 extern crate pkg_config;
 
 extern crate cc;
-extern crate glob;
 
+use std::ffi::OsStr;
 use std::path::PathBuf;
 use std::{env, fs};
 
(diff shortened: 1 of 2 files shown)


zstd-safe/zstd-sys/build.rs
```diff
@@ -89,18 +86,19 @@ fn enable_threading(_config: &mut cc::Build) {}
...
+    for dir in &[
+        "zstd/lib/common",
+        "zstd/lib/compress",
+        "zstd/lib/decompress",
+        "zstd/lib/legacy",
+        "zstd/lib/dictBuilder",
+    ] {
+        for entry in fs::read_dir(dir).unwrap() {
+            let path = entry.unwrap().path();
+            if path.extension() == Some(OsStr::new("c")) {
```
gyscos, line 99: I think we may also want to check that it's a file and not a directory.

gyscos:
Sure, and thanks for the work!

benesch:
Awesome, thanks very much!

PR #108, opened 2021-03-26
Added feature "thin" for reducing binary size

DezzK:
Added feature "thin" for reducing binary size using recommendations from zstd/lib.


gyscos:
Thanks for the work!

This looks great! Though I wonder if the rust part does not outweigh the C lib in the final binary size.

PR #109, opened 2021-03-31
Expose the accepted range of compression levels.

khuey:
Should be straightforward. The reason for adding `impl Debug` is so that in my program I can easily print the accepted range if the user passes an out-of-bounds value.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,10 +28,17 @@ pub mod block;
 pub mod dict;
 pub mod stream;
 
+use std::fmt::Debug;
 use std::io;
+use std::ops::RangeBounds;
 
 /// Default compression level.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -28,11 +28,18 @@ pub mod block;
...
 
+use std::fmt::Debug;
 use std::io;
+use std::ops::RangeBounds;
 
 /// Default compression level.
 pub use zstd_safe::CLEVEL_DEFAULT as DEFAULT_COMPRESSION_LEVEL;
 
+/// The accepted range of compression levels.
+pub fn compression_level_range() -> impl RangeBounds<zstd_safe::CompressionLevel> + Debug {
```
gyscos, line 39: Any reason not to just return a [`RangeInclusive`](https://doc.rust-lang.org/beta/std/ops/struct.RangeInclusive.html)?

khuey: Not really.

PR #113, opened 2021-04-08
Add `Container` trait for resizable bytes container

gyscos:


zstd-safe/src/lib.rs
```diff
@@ -1048,15 +1084,129 @@ pub struct InBuffer<'a> {
...
+/// Can start from uninitialized memory, and will be partially filled.
+///
+/// Should be implemented by a contiguous chunk of memory.
+///
+/// The main implementors are:
+/// * `Vec<u8>` and similar structures. These can start empty with a non-zero capacity, and they
+///   will be resized to cover the data written.
+/// * `[u8]` and `[u8; N]`. These must start already-initialized, and will not be resized. It will
+///   be up to the caller to only use the part that was written.
+pub trait WriteBuf {
```
anp, line 1098: I think this trait might need to be `unsafe` since `as_slice()` needs to uphold the initialization invariant but probably shouldn't itself be an unsafe method.

zstd-safe/src/lib.rs
```diff
@@ -1067,38 +1217,73 @@ fn ptr_mut<B>(ptr_void: &mut B) -> *mut B {
...
+    pub fn pos(&self) -> usize {
+        self.pos
+    }
+
+    /// Sets the new cursor position.
+    ///
+    /// # Panics
+    ///
+    /// If `pos > self.dst.capacity()`.
+    pub fn set_pos(&mut self, pos: usize) {
```
anp, line 1272: Does this need to be marked `unsafe`?

zstd-safe/src/lib.rs
```diff
@@ -1148,6 +1333,23 @@ impl<'a> InBuffer<'a> {
...
+    pub fn pos(&self) -> usize {
+        self.pos
+    }
+
+    /// Sets the new cursor position.
+    ///
+    /// # Panics
+    ///
+    /// If `pos > self.src.len()`.
+    pub fn set_pos(&mut self, pos: usize) {
```
anp, line 1346: If OutBuffer::set_pos should be unsafe, I think this needs to be `unsafe` too since it sets a value for calling that method on drop, right?

zstd-safe/src/lib.rs
```diff
@@ -1048,15 +1084,129 @@ pub struct InBuffer<'a> {
...
+    /// Returns a pointer to the start of the data.
+    fn as_mut_ptr(&mut self) -> *mut u8;
+
+    /// Indicates that the first `n` bytes of the container have been written.
+    unsafe fn filled_until(&mut self, n: usize);
+
+    /// Call the given closure using the pointer and capacity from `self`.
+    ///
+    /// Assumes the given function returns a parseable code, which if valid, represents how many
+    /// bytes were written to `self`.
```
anp, line 1114: I think there are a couple other invariants assumed here that would be good to document:

* `f` must not treat the memory as initialized
* the bytes written must be contiguous starting at the base of the pointer

zstd-safe/src/lib.rs
```diff
@@ -1048,15 +1084,129 @@ pub struct InBuffer<'a> {
...
+    unsafe fn write_from<F>(&mut self, f: F) -> SafeResult
+    where
+        F: FnOnce(*mut c_void, usize) -> usize,
```
anp, line 1117: What do you think about having `F: FnOnce(*mut c_void, usize) -> SafeResult` here? It'll mean typing a few more `parse_code(...)` instances where its used but it would make the contract here clearer I think.

anp (commented):
Just driving by, but this looks really promising! Have a few thoughts, mostly minor suggestions.

src/block/compressor.rs
```diff
@@ -60,13 +55,10 @@ impl Compressor {
...
-        unsafe {
-            // Use all capacity.
-            // Memory may not be initialized, but we won't read it.
-            buffer.set_len(buffer_len);
-            let len = self.compress_to_buffer(data, &mut buffer[..], level)?;
-            buffer.set_len(len);
-        }
+
+        // Use all capacity.
+        // Memory may not be initialized, but we won't read it.
```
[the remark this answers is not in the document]
gyscos: Indeed, it is now outdated.

zstd-safe/src/lib.rs
gyscos: I agree it needs to be an unsafe trait, but for different reasons: that `capacity()` must not lie about the actual length of available data.

`as_slice` itself, if implemented safely, cannot return uninitialized memory (if it could, then that uninitialized memory would probably be visible without this method). We're not _relying_ on what this implementation provides. `filled_until` is the main one that needs to be properly implemented: it's both unsafe to use _and_ to implement (maybe that should also justify unsafe trait?)

anp: Ahh, yeah this makes sense.

zstd-safe/src/lib.rs
gyscos: Indeed it should.

zstd-safe/src/lib.rs
gyscos: This is for the `InBuffer`, where we never write to (we only read from that). I don't think we set the `OutBuffer::pos` based on that either.
What we do is set the `OutBuffer::pos` based on `OutBufferWrapper::buf`, a `zstd_sys::ZSTD_outBuffer`, which is updated by the C zstd library, at the same time it actually writes to the buffer itself.

zstd-safe/src/lib.rs
gyscos: Yeaah, I was thinking about it and was lazy to type `parse_code` everywhere, but it does make it a bit clearer.

PR #115, opened 2021-04-19
Use decompression bound API to limit decompressor allocation

farnz:
There's an experimental API that gives us an upper limit on the space needed to decompress a given zstd compressed block. 

Expose this to users if the `experimental` feature is on and the block is valid, and use it in `decompress` to reduce the size of allocations if the C library promises that decompressed data is smaller than the capacity the user provided.

Without `experimental`, we always return `None` for the upper bound, which is treated by the `decompress` changes as "do not limit capacity".

### DIFF
--- a/src/block/decompressor.rs
+++ b/src/block/decompressor.rs
@@ -1,5 +1,7 @@
 use crate::map_error_code;
 
+#[cfg(feature = "experimental")]
+use std::convert::TryInto;
 use std::io;
 use zstd_safe;
 
@@ -54,7 +56,7 @@ impl Decompressor {
         capacity: usize,
     ) -> io::Result<Vec<u8>> {
         let capacity =
-            self.upper_bound(data).unwrap_or(capacity).min(capacity);
+            Self::upper_bound(data).unwrap_or(capacity).min(capacity);
         let mut buffer = Vec::with_capacity(capacity);
         unsafe {
             buffer.set_len(capacity);
@@ -66,15 +68,16 @@ impl Decompressor {
 
     /// Get an upper bound on the decompressed size of data, if available
     ///
-    /// This can be used to pre-allocate enough capacity for [`decompress_to_buffer`]
-    /// and is used by [`decompress`] to ensure that it does not over-allocate if
+    /// This can be used to pre-allocate enough capacity for `decompress_to_buffer`
+    /// and is used by `decompress` to ensure that it does not over-allocate if
     /// you supply a large `capacity`.
     ///
-    /// Will return `None` if the upper bound cannot be determined
-    pub fn upper_bound(&self, _data: &[u8]) -> Option<usize> {
+    /// Will return `None` if the upper bound cannot be determined or is larger than `usize::MAX`
+    pub fn upper_bound(_data: &[u8]) -> Option<usize> {
         #[cfg(feature = "experimental")]
         {
-            zstd_safe::decompress_bound(_data)
+            let bound = zstd_safe::decompress_bound(_data).ok()?;
+            bound.try_into().ok()
         }
         #[cfg(not(feature = "experimental"))]
         {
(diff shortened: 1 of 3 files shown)


src/block/decompressor.rs
```diff
@@ -61,6 +63,24 @@ impl Decompressor {
...
     }
+
+    /// Get an upper bound on the decompressed size of data, if available
+    ///
+    /// This can be used to pre-allocate enough capacity for [`decompress_to_buffer`]
+    /// and is used by [`decompress`] to ensure that it does not over-allocate if
+    /// you supply a large `capacity`.
+    ///
+    /// Will return `None` if the upper bound cannot be determined
+    pub fn upper_bound(&self, _data: &[u8]) -> Option<usize> {
```
gyscos, line 74: I think we can return a `SafeResult` here, to include the error code?

zstd-safe/src/lib.rs
```diff
@@ -1622,3 +1622,16 @@ pub fn decompress_block(dctx: &mut DCtx, dst: &mut [u8], src: &[u8]) -> usize {
 pub fn insert_block(dctx: &mut DCtx, block: &[u8]) -> usize {
     dctx.insert_block(block)
 }
+
+/// Wraps the `ZSTD_decompressBound` function
+#[cfg(feature = "experimental")]
+pub fn decompress_bound(data: &[u8]) -> Option<usize> {
```
gyscos, line 1628: I think we can return `SafeResult` here, to forward the error code?

gyscos: On the other hand, do we want to return a `Result<u64, ErrorCode>`?

farnz: My thinking here:

1. `Option`, not `Result`, because the only permissible errors here mean "a useful upper bound cannot be determined", and are not otherwise helpful - you will get a real error later if you try to decompress.
2. `usize`, not `u64`, because the point of the return value is to give you input to an allocator (e.g. `Vec::with_capacity`).

I am ambivalent about `Result` versus `Option`, and I'm happy to switch if you'd prefer a `SafeResult`; I don't see a good use for this function that doesn't entail a cast to `usize` somewhere, so would like a strong argument against using `usize` (since most/all use sites will want a `usize`, not a `u64`).

gyscos: There is a single failure cause _right now_ - but
* It's very possible that other error causes may be added later, for instance if some frame options prevent estimating the bounds.
* Many other functions also may have a single error cause right now; returning a `Result` helps unify their processing.

Using the result to index in a `Vec` is indeed a common option, but not the only one. In particular, it may be desirable to process files over 4GB on a 32-bit system (either writing straight to disk, or to the network, or straight to another compression stream). Especially in the `zstd-safe` library, I'd try to stay as close as possible to the C lib in this kind of guarantees and limitations. For reference, the existing `get_frame_content_size` returns a `u64`.

farnz: OK - I'll redo `zstd-safe` to use `Result<u64, ErrorCode>`, and do the conversions in the main crate directly.

gyscos: Thanks! In particular checking the error _after_ converting to `usize` sounds like it might have causes issues on 32-bits.

gyscos: Though it doesn't look like there's a method that can actually parse an error from a `u long long`? :^S Why, zstd, why?

I guess -2 in `u long long`, when truncated to `size_t`, is still -2? Which sorta kinda works? Duh, C...

farnz: > I guess -2 in `u long long`, when truncated to `size_t`, is still -2? Which sorta kinda works? Duh, C...

C is, in my long experience, a great language for confusing people into seeing bugs that don't exist, and not seeing the bugs that do exist.

Hence my shift to Rust :-)

gyscos:
Looks good - thanks for the work!

PR #116, opened 2021-04-28
Exclude unneeded files from the C repo when packaging

Jake-Shadle:
I happened to be watching the output of a fresh fetch and noticed that zstd-sys was the largest crate that was downloaded at 1.9MiB and suspected it was due to including the entirety of the C lib's repo, which was indeed the case. This just excludes everything from that repo that isn't needed for actually compiling the lib, and the the LICENSE/COPYING file of course.

This makes the crate go from 1.9MiB to 632KiB, so a little less than 1/3 of the current size.


gyscos:
Indeed - thanks for the work!

Jake-Shadle:
Cool, thanks for the quick merge!

PR #120, opened 2021-05-15
Fix the Decoder API sometimes not outputting all available data

### LINKED ISSUE #119: `Decoder` API doesn't always output all available data if the input stream blocks
The first thing a read does is call `fill_buf`:

https://github.com/gyscos/zstd-rs/blob/6713e1dd79cb4b0bdb9970825d984245034262a5/src/stream/zio/reader.rs#L97

If all available data has already been passed to zstd in a prior call, this call might block (it might e.g. be a TCP stream, as in my case). zstd sometimes has its own buffer so that a call with an empty input buffer would yield uncompressed bytes.

heinrich5991:
This is done by first asking zstd for more data, before trying to read
more compressed bytes from the data source.

An alternative would be the
[`BufReader::buffer`](https://doc.rust-lang.org/std/io/struct.BufReader.html#method.buffer)
function, but that is not available because we only have a type that
satisfies `BufRead` in this context.

I tested this with my TCP connection and it solved my problem.

Fixes #119.

I haven't tested this for performance regressions.

### DIFF
--- a/src/stream/zio/reader.rs
+++ b/src/stream/zio/reader.rs
@@ -93,7 +93,7 @@ where
         loop {
             let (bytes_read, bytes_written) = {
                 // Start with a fresh pool of un-processed data.
-                // This is the only line that can return an interuption error.
+                // This is the only line that can return an interruption error.
                 let input = fill_buf(&mut self.reader)?;
 
                 // println!("{:?}", input);
--- a/src/stream/zio/reader.rs
+++ b/src/stream/zio/reader.rs
@@ -90,17 +90,19 @@ where
         }
 
         // Keep trying until _something_ has been written.
+        let mut first = true;
         loop {
             let (bytes_read, bytes_written) = {
                 // Start with a fresh pool of un-processed data.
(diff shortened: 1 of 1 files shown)


heinrich5991:
Could you approve the workflows? :)

gyscos:
Thanks! Looks good.
I am tempted to force the use of a `BufReader` instead of just the `BufRead` trait, to avoid a potential extra call into the zstd library everytime, but that may come later (and hopefully by then informed by benchmarks).

gyscos:
Hmm turns out this is causing an issue with the reader-decoder. It seems to trigger an "incomplete frame" error.
Will investigate a bit to see if there's an easy fix, but I may end up reverting back for now.

heinrich5991:
Thanks for fixing my fix. :)

Would you mind creating a crates.io release with this bug fixed? I could drop my local patch then.

gyscos:
0.8.2 has been published

heinrich5991:
Thanks a lot. :)

PR #123, opened 2021-06-07
Cargo bug workaround

olivierlemasle:
Using `cargo vendor` with `zstd-sys` does not copy the `zstd` directory, due to a bug in Cargo: https://github.com/rust-lang/cargo/issues/9555

This commit replaces the "exclude" field with a negation, which causes an issue with current version of Cargo, by an equivalent
"include" field.

### DIFF
--- a/zstd-safe/zstd-sys/Cargo.toml
+++ b/zstd-safe/zstd-sys/Cargo.toml
@@ -19,13 +19,24 @@ repository = "https://github.com/gyscos/zstd-rs"
 version = "1.6.0+zstd.1.5.0"
 edition = "2018"
 
-exclude = [
-    "zstd",
-    "!zstd/LICENSE",
-    "!zstd/COPYING",
-    "!zstd/lib/**/**.h",
-    "!zstd/lib/**/**.c",
+# Use include instead of exclude, as a (temporary)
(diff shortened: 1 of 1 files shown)


zstd-safe/zstd-sys/Cargo.toml
```diff
@@ -19,13 +19,24 @@ repository = "https://github.com/gyscos/zstd-rs"
...
-    "!zstd/LICENSE",
-    "!zstd/COPYING",
-    "!zstd/lib/**/**.h",
-    "!zstd/lib/**/**.c",
+# Use include instead of exclude, as a (temporary)
+# workaround for https://github.com/rust-lang/cargo/issues/9555
+include = [
+    "/LICENSE",
+    "/*.*",
+    "/src/**/*",
```
gyscos, line 27: Could we just specify `/src/` here?

olivierlemasle: Yes, it will include the same files.

olivierlemasle: I've updated the PR.

gyscos:
Thanks for the work! I guess we'll need to stick for a while with this "temporary" change, since all current versions of cargo are affected.

PR #132, opened 2021-09-03
Generic closures for AutoFinishEncoder for deriving autotraits

l4l:
I'm not sure, but still tried to make it API non-breaking change.

The patch uses generic for `AutoFinishEncoder` closure, so it will be possible to derive marker traits for the type itself. In particularly `Send` trait, as presented in the test.

### DIFF
--- a/src/stream/write/mod.rs
+++ b/src/stream/write/mod.rs
@@ -41,22 +41,25 @@ pub struct Decoder<'a, W: Write> {
 ///
 /// [`auto_finish()`]: Encoder::auto_finish
 /// [`Encoder`]: Encoder
-pub struct AutoFinishEncoder<'a, W: Write> {
+pub struct AutoFinishEncoder<'a, W: Write, F = Box<dyn FnMut(io::Result<W>)>>
+where
+    F: FnMut(io::Result<W>) + 'static,
+{
     // We wrap this in an option to take it during drop.
     encoder: Option<Encoder<'a, W>>,
 
     // TODO: make this a FnOnce once it works in a Box
-    on_finish: Option<Box<dyn FnMut(io::Result<W>)>>,
+    on_finish: Option<F>,
 }
 
-impl<'a, W: Write> AutoFinishEncoder<'a, W> {
-    fn new<F>(encoder: Encoder<'a, W>, on_finish: F) -> Self
-    where
-        F: 'static + FnMut(io::Result<W>),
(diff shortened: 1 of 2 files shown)


src/stream/write/mod.rs
```diff
@@ -41,22 +41,25 @@ pub struct Decoder<'a, W: Write> {
...
 }
 
-impl<'a, W: Write> AutoFinishEncoder<'a, W> {
-    fn new<F>(encoder: Encoder<'a, W>, on_finish: F) -> Self
-    where
-        F: 'static + FnMut(io::Result<W>),
-    {
+impl<'a, W: Write, F> AutoFinishEncoder<'a, W, F>
+where
+    F: FnMut(io::Result<W>) + 'static,
```
gyscos, line 57: I don't think `'static` is still required, now that `F` is part of `Self` itself.

src/stream/write/mod.rs
```diff
@@ -41,22 +41,25 @@ pub struct Decoder<'a, W: Write> {
 ///
 /// [`auto_finish()`]: Encoder::auto_finish
 /// [`Encoder`]: Encoder
-pub struct AutoFinishEncoder<'a, W: Write> {
+pub struct AutoFinishEncoder<'a, W: Write, F = Box<dyn FnMut(io::Result<W>)>>
```
gyscos, line 44: Wondering if we could make `F` include `Send` by default as well?

src/stream/write/mod.rs
```diff
@@ -74,7 +77,10 @@ impl<'a, W: Write> AutoFinishEncoder<'a, W> {
     }
 }
 
-impl<W: Write> Drop for AutoFinishEncoder<'_, W> {
+impl<W: Write, F> Drop for AutoFinishEncoder<'_, W, F>
+where
+    F: FnMut(io::Result<W>) + 'static,
```
gyscos, line 82: Here too, `'static` might not be required?

gyscos (commented):
Thanks for the work, this sounds like a good idea!

Just a few minor comments.

I think you'll also want to update other implementations like `impl<W: Write> Write for AutoFinishEncoder<'_, W>` (which currently only apply to the default parameter).

gyscos:
I ended up implementing this in the latest commit, to go with the new `AutoFlushDecoder`.

Thanks again for the work!

PR #133, opened 2021-09-19
Fix crates.io badge and update links

atouchet:


gyscos:
Hi! Looks great, thanks for the pr!

PR #139, opened 2022-01-26
zstd-sys: impl wasm32-unknown-unknown support

quininer:
This PR adds a simple wasm shim to `zstd-sys` to support `wasm32-unknown-unknown` arch.

The `legacy` and `dict_builder` cannot be compiled on this shim because they depend on `stdio.h`, so I added a new feature cfg to turning off `dict_builder` code.

You can test it with `wasmtime`

```
> cd zstd-safe/zstd-sys/
> cargo build --target wasm32-unknown-unknown --no-default-features --example it_work
> wasmtime target/wasm32-unknown-unknown/debug/examples/it_work.wasm --invoke test_compress
```

### DIFF
--- a/examples/train.rs
+++ b/examples/train.rs
@@ -30,11 +30,8 @@ fn main() {
 
     let files: Vec<_> = matches.values_of("FILE").unwrap().collect();
 
-    #[cfg(feature = "zdict_builder")]
-    {
-        let dict = zstd::dict::from_files(&files, size).unwrap();
+    let dict = zstd::dict::from_files(&files, size).unwrap();
 
-        let mut dict_reader: &[u8] = &dict;
-        io::copy(&mut dict_reader, &mut io::stdout()).unwrap();
-    }
+    let mut dict_reader: &[u8] = &dict;
+    io::copy(&mut dict_reader, &mut io::stdout()).unwrap();
 }
--- a/examples/train.rs
+++ b/examples/train.rs
@@ -30,8 +30,11 @@ fn main() {
 
     let files: Vec<_> = matches.values_of("FILE").unwrap().collect();
 
-    let dict = zstd::dict::from_files(&files, size).unwrap();
+    #[cfg(feature = "zdict_builder")]
+    {
+        let dict = zstd::dict::from_files(&files, size).unwrap();
 
-    let mut dict_reader: &[u8] = &dict;
-    io::copy(&mut dict_reader, &mut io::stdout()).unwrap();
+        let mut dict_reader: &[u8] = &dict;
+        io::copy(&mut dict_reader, &mut io::stdout()).unwrap();
+    }
 }
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -6,7 +6,7 @@ use std::{env, fs};
 fn generate_bindings(defs: Vec<&str>, headerpaths: Vec<PathBuf>) {
     let bindings = bindgen::Builder::default()
         .header("zstd.h");
-    #[cfg(feature = "zdict")]
+    #[cfg(feature = "zdict_builder")]
     let bindings = bindings.header("zdict.h");
     let bindings = bindings
         .blocklist_type("max_align_t")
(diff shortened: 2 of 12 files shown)


gyscos:
Hi, and thanks for the work!

I think without `dictBuilder`, the functions from `zdict` don't have a definition anymore, and should not be used.

This means that the bindings file needs to depend on this. In practice, the `zstd.h` file used to generate the bindings should not include `zdict.h` when the dict builder feature is not enabled.

Rather than adding yet another dimension to the selection of the bindings file, we could split this bindings into the main zstd part and the dict part, and include each file conditionally. I pushed something like that to the [bindings_split](https://github.com/gyscos/zstd-rs/tree/bindings_split/zstd-safe/zstd-sys) branch if you want to have a look.
If it looks fine, I'll merge the split, and rebase this PR to _not_ include the zdict bindings without the feature.

quininer:
Yes, it would be nice to split them. Thanks for your quick work.

zstd-safe/zstd-sys/build.rs
```diff
@@ -5,8 +5,10 @@ use std::{env, fs};
 #[cfg(feature = "bindgen")]
 fn generate_bindings(defs: Vec<&str>, headerpaths: Vec<PathBuf>) {
     let bindings = bindgen::Builder::default()
-        .header("zstd.h")
-        .header("zdict.h")
+        .header("zstd.h");
+    #[cfg(feature = "zdict")]
+    let bindings = bindings.header("zdict.h");
+    bindings
```
gyscos, line 11:
```suggestion
    let bindings = bindings
```
Otherwise it will consume `bindings` (and probably not compile) and not actually update it in-place.

zstd-safe/zstd-sys/build.rs
```diff
@@ -19,9 +21,10 @@ fn generate_bindings(defs: Vec<&str>, headerpaths: Vec<PathBuf>) {
         .clang_args(defs.into_iter().map(|def| format!("-D{}", def)));
 
     #[cfg(feature = "experimental")]
-    let bindings = bindings
-        .clang_arg("-DZSTD_STATIC_LINKING_ONLY")
-        .clang_arg("-DZDICT_STATIC_LINKING_ONLY");
+    let bindings = bindings.clang_arg("-DZSTD_STATIC_LINKING_ONLY");
+    #[cfg(feature = "experimental")]
+    #[cfg(feature = "zdict")]
```
gyscos, line 26:
```suggestion
    #[cfg(all(feature = "experimental", feature = "zdict"))]
```

zstd-safe/zstd-sys/build.rs
```diff
@@ -114,6 +118,14 @@ fn compile_zstd() {
         config.file("zstd/lib/decompress/huf_decompress_amd64.S");
     }
 
+    if env::var("CARGO_CFG_TARGET_ARCH").ok() == Some("wasm32".into()) {
```
gyscos, line 121: Since we're re-checking this further down, I would save it as a variable.

```suggestion
    let is_wasm32 = env::var("CARGO_CFG_TARGET_ARCH").ok() == Some("wasm32".into());
    if is_wasm32 {
```

gyscos: Also, do we need that for all wasm32 targets? What about emscripten - doesn't it define its own stdio and such?

zstd-safe/zstd-sys/build.rs
```diff
@@ -152,7 +166,9 @@ fn compile_zstd() {
      * 7+: events at every position (*very* verbose)
      */
     #[cfg(feature = "debug")]
-    config.define("DEBUGLEVEL", Some("5"));
+    if is_wasm_unknown_unknown {
```
gyscos, line 169: I think you want the opposite here?
```suggestion
    if !is_wasm_unknown_unknown {
```

gyscos:
Looking good! One last thing: we'll want to add `dict_builder` features to the `zstd-safe` and `zstd` crates, and forward it downstream.

We'll also want to conditionally hide the relevant types in these crates depending on this feature.

examples/train.rs
```diff
@@ -30,8 +30,11 @@ fn main() {
...
-    let dict = zstd::dict::from_files(&files, size).unwrap();
+    #[cfg(feature = "zdict_builder")]
+    {
+        let dict = zstd::dict::from_files(&files, size).unwrap();
 
-    let mut dict_reader: &[u8] = &dict;
-    io::copy(&mut dict_reader, &mut io::stdout()).unwrap();
+        let mut dict_reader: &[u8] = &dict;
+        io::copy(&mut dict_reader, &mut io::stdout()).unwrap();
+    }
```
gyscos, line 39: You can specify that an example requires some features. In `Cargo.toml`:

```toml
[[example]]
name = "train"
required-features = ["zdict_builder"]
```

gyscos:
Thanks again for the work!

PR #145, opened 2022-03-15
Include wasm-shim files in cargo package

kylebarron:
Building the released `0.11` does not work for `wasm32-unknown-unknown`, while the build does work for the same revision from Git. 

In the process of making an example (https://github.com/kylebarron/zstd-rs-wasm-bug-example), I realized the shim files were likely not included in the release. `cargo package --list` in `zstd-sys` confirmed that previously the `wasm-shim` files were not included in the build.

### DIFF
--- a/zstd-safe/zstd-sys/Cargo.toml
+++ b/zstd-safe/zstd-sys/Cargo.toml
@@ -25,12 +25,12 @@ include = [
     "/LICENSE",
     "/*.*",
     "/src/",
+    "/wasm-shim/**/*.h",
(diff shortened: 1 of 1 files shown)


gyscos:
Agggh cargo package...
Thanks for the fix!

kylebarron:
Are the failing tests on `master` due to this PR? Looks like they might have been failing previously?

gyscos:
Yeah it's just the windows VM using outdated GCC, and my hack to update it in CI isn't fully working.

PR #153, opened 2022-05-16
Add multithread configuration to Decoder

steviez:
This was a copy/paste of the function in Encoder, with changing `compression` to `decompression` in the doc comments


gyscos:
Hi, and thanks for the work!

Note that `zstd_safe::CParameter::NbWorkers` is a compression parameter. There doesn't seem to be a decompression equivalent in the C library. I suspect that parallel decompression of a single stream is not supported at all.
(Of course if you process many streams, each can be processed in parallel.)

PR #155, opened 2022-07-03
Unconditionally enable thin LTO in for `zstd-sys`

NobodyXu:
Fat LTO usually can reduce binary size while improving the performance.
Thin LTO does the same, but taking less time.

Signed-off-by: Jiahao XU [Jiahao_XU@outlook.com](mailto:Jiahao_XU@outlook.com)

### DIFF
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -135,12 +135,13 @@ fn compile_zstd() {
 
     config.define("ZSTD_LIB_DEPRECATED", Some("0"));
 
+    config.flag_if_supported("-flto=thin");
+
     #[cfg(feature = "thin")]
     {
         config.define("HUF_FORCE_DECOMPRESS_X1", Some("1"));
         config.define("ZSTD_FORCE_DECOMPRESS_SEQUENCES_SHORT", Some("1"));
         config.define("ZSTD_NO_INLINE ", Some("1"));
-        config.flag_if_supported("-flto=thin");
         config.flag_if_supported("-Oz");
     }
 


gyscos:
Thanks for the work!

syncom:
The change in this PR, for a reason I don't understand, seems to break some statically linked cross builds having `zstd` as a dependency.
Repro steps:
1. Checkout this branch: https://github.com/syncom/rust-cross-build-nix/tree/syncom/poc/zstd-rs-breaking-commit, which [pins](https://github.com/syncom/rust-cross-build-nix/blob/32c3a64ff39bee92602c8c425291c40bb7c5a6f4/Cargo.toml#L10) `zstd-rs` at the revision that introduces this change.
2. Install [cross](https://github.com/cross-rs/cross).
3. Run command `cross build --target=aarch64-unknown-linux-musl`.  I got the following error during linking:
[aarch64.cross-error.txt](https://github.com/gyscos/zstd-rs/files/9544105/aarch64.cross-error.txt). Command `cross build --target=x86_64-unknown-linux-musl` renders a similar failed build.
4. Reverting this commit, I was able to perform the above cross builds successfully, as shown [here](https://github.com/syncom/rust-cross-build-nix/blob/33005fbcd4a1be0f604afbf1415b51004dda1d70/Cargo.toml#L12), either with `cross` or with Nix.

I was wondering if we should have this change reverted until the root cause for this is known, if the above repro steps are valid.

NobodyXu:
Could that be a `cross` problem?

I run `cargo b --target x86_64-unknown-linux-musl` and it just worked.
`cross b --target x86_64-unknown-linux-musl` indeed failed, but since `cross` has a history of failing and requires workaround, I'm not surprised if it's `cross` being the culprit again.

syncom:
> Could that be a `cross` problem?

It could be. I don't understand this failure very well. I first noticed this breakage when trying to build in Nix with [rust-overlay](https://github.com/oxalica/rust-overlay). Building for `aarch64-unknown-linux-musl` fails, but it succeeds for `x86_64-unknown-linux-musl` with [the methodology](https://github.com/syncom/rust-cross-build-nix/tree/main/nix-rust-overlay)

PR #159, opened 2022-07-20
Add new feature flag "static"

### LINKED ISSUE #158: Add a feature flag "static" to ensure `libzstd` is statically linked
In [`cargo-binstall`](https://github.com/ryankurte/cargo-binstall), we need to statically link with `libzstd` to create a portable binary.

NobodyXu:
Resolved #158 


NobodyXu:
Not sure why the CI for windows i686 failed.

zstd-safe/zstd-sys/build.rs
```diff
@@ -200,7 +200,9 @@ fn main() {
     }
 
     // println!("cargo:rustc-link-lib=zstd");
-    let (defs, headerpaths) = if cfg!(feature = "pkg-config") {
+    let (defs, headerpaths) = if cfg!(feature = "pkg-config")
+        && !cfg!(feature = "static")
```
gyscos, line 204: I think we should make it an error to have both pkg-config and static features, rather than silently dropping one.

NobodyXu: I prefer `static` to override `pkg-config`, since it is possible for one dependency to request `pkg-config` and another to request `static`.

In this situation, it is not always possible to modify or it maybe hard to remove that and I would prefer statically linked to dynamically linked zstd.

gyscos:
Note that static linking is already the default.
I believe the only way to build dynamically right now is using the `pkg-config` feature, which is a conscious decision.

* If you're not using this feature right now, then you are statically linking already.
* If you _are_ using this feature right now and it leads to dynamic linking, then _something_ explicitly asked for dynamic linking. It would be safer to find where this came from and fix that, than to silently disregard this request.

The main reason to add a static feature (and it may be a valid one!) is to detect if the situation ever changes: if some dependency starts requiring `pkg-config`, rather than silently switching to dynamic linking, it would at least block compilation to make the change visible.

Also note that pkg-config does not _have_ to result in dynamic linking. It will _try_ static linking if the `libzstd.a` file is available, but if it's not - for example the archlinux zstd package does not ship static libraries - it will fallback to dynamic linking.

In that case, it may be better to detect when pkg-config fails to enable static linking ([it uses a private function to check that](https://docs.rs/pkg-config/latest/src/pkg_config/lib.rs.html#629)), and only _then_ trigger a build failure.
Now - how to detect this failure? We can either re-implement this function locally? Talk to the pkg_config maintainers and make it public? There doesn't seem to be an easy-to-use field in the resulting `Library` that would indicate this. :^S

zstd-safe/zstd-sys/build.rs
gyscos: If a dependency asks for pkg-config, we should honor this decision and not discard it.

Though in general asking for `pkg-config` is not something re-usable libraries should be doing, it's more of a end-of-the-line binary decision.

NobodyXu:
@gyscos I just realized that `pkg_config` function in `build.rs` set [`pkg_config::Config::statik`](https://docs.rs/pkg-config/latest/pkg_config/struct.Config.html#method.statik) to `true`, so maybe we don't need this new feature afterall?

gyscos:
> Also note that pkg-config does not have to result in dynamic linking. It will try static linking if the libzstd.a file is available, but if it's not - for example the archlinux zstd package does not ship static libraries - it will fallback to dynamic linking.

It's an unfortunate decision, as it does not appear to make this very visible to the caller.

PR #162, opened 2022-08-31
Deterministic file adding order in compile_zstd()

syncom:
This is to make the build more deterministic.

Note that `fs::read_dir()`'s behavior is platform-specific. We sort the
returned result of `read_dir` to make the file addition order
deterministic.

### DIFF
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -92,10 +92,9 @@ fn compile_zstd() {
         #[cfg(feature = "legacy")]
         "zstd/lib/legacy",
     ] {
-        let mut entries: Vec<_> = fs::read_dir(dir).unwrap().map(|r| r.unwrap()).collect();
-        entries.sort_by_key(|k| k.path());
-        for entry in entries {
-            let path = entry.path();
+        let mut entries: Vec<_> = fs::read_dir(dir).unwrap().map(|r| r.unwrap().path()).collect();
+        entries.sort();
+        for path in entries {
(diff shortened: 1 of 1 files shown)


zstd-safe/zstd-sys/build.rs
```diff
@@ -92,8 +92,10 @@ fn compile_zstd() {
         #[cfg(feature = "legacy")]
         "zstd/lib/legacy",
     ] {
-        for entry in fs::read_dir(dir).unwrap() {
-            let path = entry.unwrap().path();
+        let mut entries: Vec<_> = fs::read_dir(dir).unwrap().map(|r| r.unwrap()).collect();
+        entries.sort_by_key(|k| k.path());
```
gyscos, line 96: Note: `DirEntry::path()` actually creates a new `PathBuf` everytime.

Since we only care about these paths, it may be easier to collect the Paths directly.

zstd-safe/zstd-sys/build.rs
```diff
@@ -92,8 +92,10 @@ fn compile_zstd() {
...
+        for entry in entries {
+            let path = entry.path();
```
gyscos, line 98:
```suggestion
        let mut entries: Vec<_> = fs::read_dir(dir).unwrap().map(|r| r.unwrap().path()).collect();
        entries.sort();
        for path in entries {
```

bmwiedemann:
Hi, could you please make a release with this fix, so that it is easier to include in our openSUSE vendor tarballs?

PR #163, opened 2022-08-31
Disable unused features of dep `bindgen`

NobodyXu:
`bindgen` enables feature "clap" and "logging" by default, which is used only in binary but not in library.


NobodyXu:
The CI also failed on master, so I think this PR itself is ok.

gyscos:
Thanks for the work!

PR #167, opened 2022-10-04
Disable -flto=thin on GNU compilers

gyscos:


zstd-safe/zstd-sys/build.rs
```diff
@@ -138,7 +138,11 @@ fn compile_zstd() {
 
     config.define("ZSTD_LIB_DEPRECATED", Some("0"));
 
-    config.flag_if_supported("-flto=thin");
+    if !config.get_compiler().is_like_gnu() {
+        // gcc has a -flto but not -flto=thin
+        // Apparently this is causing crashes on windows-gnu?
```
NobodyXu, line 143: Maybe we can enable fat LTO for windows?

gyscos: The PR doesn't even fix windows build, so I'll need to find something else :(

PR #172, opened 2022-10-13
Improve zstd-sys/build.rs: Fallback to `-Os` or `-O2` if `-Oz` is not supported

NobodyXu:

### DIFF
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -78,6 +78,18 @@ fn enable_threading(config: &mut cc::Build) {
 #[cfg(not(feature = "zstdmt"))]
 fn enable_threading(_config: &mut cc::Build) {}
 
+/// This function would find the first flag in `flags` that is supported
+/// and add that to `config`.
+fn flag_if_supported_with_fallbacks(config: &mut cc::Build, flags: &[&str]) {
+    let option = flags
+        .iter()
+        .find(|flag| config.is_flag_supported(flag).unwrap_or_default());
+
+    if let Some(flag) = option {
+        config.flag(flag);
+    }
+}
+
 fn compile_zstd() {
     let mut config = cc::Build::new();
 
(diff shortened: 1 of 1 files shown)


NobodyXu:
@gyscos Can you review this PR please?

NobodyXu:
@gyscos Pinging

gyscos:
Unfortunately I'm not sure `is_flag_supported` does its job: it seems that for some versions of gcc, it will accept a flag, even though this flag leads to a non-working compilation.

I'm still trying to figure out how to safely enable any form of lto on GCC that will work on both recent gcc version, older gcc version used in cross (gcc 9), and the gcc used in the windows-gnu target...

NobodyXu:
> Unfortunately I'm not sure `is_flag_supported` does its job: it seems that for some versions of gcc, it will accept a flag, even though this flag leads to a non-working compilation.
> 
> I'm still trying to figure out how to safely enable any form of lto on GCC that will work on both recent gcc version, older gcc version used in cross (gcc 9), and the gcc used in the windows-gnu target...

Perhaps we should just compile a simple hello-world with lto to see if that works?
If it can be compiled, then it's likely it would work.

I can try implementing that in this PR.

gyscos:
Actually, do we need to turn on O2 manually? What's the advantage vs letting cc pick the optimization level?

NobodyXu:
> Actually, do we need to turn on O2 manually? What's the advantage vs letting cc pick the optimization level?

It's only turned on if feature "thin" is enabled.
Using -O2 probably generates smaller code than -O3 (or even -O1 sometimes).

NobodyXu:
> I realize it's only when none of Oz or Os works (not sure when that happens),

I remember that gcc doesn't support -Oz and only -Os?

NobodyXu:
> Unfortunately I'm not sure `is_flag_supported` does its job: it seems that for some versions of gcc, it will accept a flag, even though this flag leads to a non-working compilation.
> 
> I'm still trying to figure out how to safely enable any form of lto on GCC that will work on both recent gcc version, older gcc version used in cross (gcc 9), and the gcc used in the windows-gnu target...

I checked [`Build::is_flag_supported`](https://docs.rs/cc/latest/src/cc/lib.rs.html#488-544) and it actually attempts to compile the file with the flag provided.

I think the problem lies in [this line](https://docs.rs/cc/latest/src/cc/lib.rs.html#540), where it does not check the exit status of the command and assumes that if the stderr is empty, everything works as expected.

NobodyXu:
@gyscos I've opend https://github.com/rust-lang/cc-rs/pull/757 which might help fix the LTO issue in zstd.

PR #175, opened 2022-11-03
add environment variable to force usage of pkg-config

### LINKED ISSUE #150: Consider adding an environment variable for dynamic linking
Would you be ok with an environment variable that could be used to specify dynamic linking ?
At the moment dynamic linking relies on the "pkg-config" feature.
having the condition include also an environment variable would 
- make it easy to use dynamic linking in a dev environment (even for dependencies of dependencies). It could save precious compile time
- Make it easy for some linux distributions to avoid static compilation and rely on dynamic linking (where static compilation fails for some reason, you could rely on a pre-built version, (one example of this is nixos).

if you are ok with this, I would be happy to make a PR.
how about the name CARGO_ZSTD_SYS_DYNAMIC_LINKING for the environment variable ?

Last but not least, thanks a lot for the great library!

figsoda:
fixes #150 

this makes pkg-config a non-optional build-dependency

now having `ZSTD_SYS_USE_PKG_CONFIG` will be equivalent to having the pkg-config feature

### DIFF
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -188,6 +188,7 @@ fn compile_zstd() {
 }
 
 fn main() {
+    #[cfg(not(feature = "non-cargo"))]
     println!("cargo:rerun-if-env-changed=ZSTD_SYS_USE_PKG_CONFIG");
 
     let target_arch =
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -38,7 +38,6 @@ fn generate_bindings(defs: Vec<&str>, headerpaths: Vec<PathBuf>) {
 #[cfg(not(feature = "bindgen"))]
 fn generate_bindings(_: Vec<&str>, _: Vec<PathBuf>) {}
 
-#[cfg(feature = "pkg-config")]
 fn pkg_config() -> (Vec<&'static str>, Vec<PathBuf>) {
     let library = pkg_config::Config::new()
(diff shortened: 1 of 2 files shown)


gyscos:
Unfortunately I'm not sure `is_flag_supported` does its job: it seems that for some versions of gcc, it will accept a flag, even though this flag leads to a non-working compilation.

I'm still trying to figure out how to safely enable any form of lto on GCC that will work on both recent gcc version, older gcc version used in cross (gcc 9), and the gcc used in the windows-gnu target...

zstd-safe/zstd-sys/build.rs
```diff
@@ -194,6 +188,8 @@ fn compile_zstd() {
 }
 
 fn main() {
+    println!("cargo:rerun-if-env-changed=ZSTD_SYS_USE_PKG_CONFIG");
```
gyscos, line 191: Let's only run this if the `non-cargo` feature is not set.

zstd-safe/zstd-sys/build.rs
```diff
@@ -203,7 +199,9 @@ fn main() {
     }
 
     // println!("cargo:rustc-link-lib=zstd");
-    let (defs, headerpaths) = if cfg!(feature = "pkg-config") {
+    let (defs, headerpaths) = if cfg!(feature = "pkg-config")
```
gyscos, line 202: We may need to add `pkg-config = []` to the features array in `Cargo.toml` to allow setting this as a feature.

gyscos (approved):

PR #176, opened 2022-11-21
zstd-sys: Don't use `target*` cfg in `build.rs` to fix Windows cross-compile

MarijnS95:
This issue got exposed with [CC upgrading to 1.0.77] and now detecting the `.S` file extension to pass on to `ml64.exe` (MASM) on Windows; we don't have that on our cross-compiling (from Linux to Windows) CI setup, hence uncovered that this build script was using the _`target_os` flag from the host_ in the build script to make a decision on _what to compile for the target_.

[CC upgrading to 1.0.77]: https://github.com/rust-lang/cc-rs/compare/1.0.76...1.0.77

---

Curiously this implies the `.S` file must have compiled fine for us on `cc 1.0.76` and prior, or somehow got skipped (expecting errors when `ZSTD_DISABLE_ASM` isn't defined...).

### DIFF
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -114,7 +114,7 @@ fn compile_zstd() {
 
     // Either include ASM files, or disable ASM entirely.
     // Also disable it on windows, apparently it doesn't do well with these .S files at the moment.
-    if cfg!(any(target_os = "windows", feature = "no_asm")) {
+    if cfg!(feature = "no_asm") || std::env::var("CARGO_CFG_WINDOWS").is_ok() {
         config.define("ZSTD_DISABLE_ASM", Some(""));
     } else {
         config.file("zstd/lib/decompress/huf_decompress_amd64.S");


gyscos:
Indeed, thanks for identifying and fixing this issue!

gyscos:
Indeed as found in https://github.com/rust-lang/cc-rs/issues/754 the compiler was previously ignoring `.s` files, only looking at `.asm`.

Hainish:
Is it possible to make a bugfix release with this fix? This library lies upstream of a cross-compiled project I'd like to make a release for.

gyscos:
Just published zstd-sys 2.0.2.

MarijnS95:
As per https://github.com/rust-lang/cc-rs/issues/754#issuecomment-1325702826 I am in hindsight not entirely sure if this is the right solution: since it was compiling just fine before in a cross-compilation setup, were we perhaps having an advantage because `clang-cl` is available and could compile the ASM even for Windows targets?

If that's true, perhaps we should instead check/force the compiler to be used when cross-compiling, instead of disabling it outright because of expecting `ml64.exe` to be used.

If anything [`cc` just backed out of the change](https://github.com/rust-lang/cc-rs/pull/755) and now uses `clang-cl` again for `.S` files (although the community doesn't seem to have agreed yet on that distinction) like it did before.

PR #183, opened 2023-01-07
Add support for `wasm32-wasi`

VivekPanyam:
This PR enables building with the `wasm32-wasi` target by using the same shims that are used for the `wasm32-unknown-unknown` target.

Both of the following commands run successfully
```
cargo build  --target wasm32-unknown-unknown --no-default-features
cargo build  --target wasm32-wasi --no-default-features
```


gyscos:
Thanks for the PR!

Issue #13, opened 2016-09-04
Skippable frame compatibility


gyscos:
[Skippable frames](https://github.com/facebook/zstd/wiki/Zstandard-Compression-Format#skippable-frames) is a feature in the Zstd format, allowing user data to be inserted between zstd frames.

This means when decoding concatenated frames, we should skip those frames.

PR #185, opened 2023-01-19
Add support for skippable frames

antoyo:
Hi.
This add supports for writing and reading skippable frames as well as skipping frames.

There is no support for doing that in a streaming way in the ZSTD C library, so that's why so part of this PR parses some stuff manually.

There are a few TODOs in the code that I'd like you to tell what you think we should do with that.

Thanks for the review.

### DIFF
--- a/src/stream/zio/reader.rs
+++ b/src/stream/zio/reader.rs
@@ -47,6 +47,11 @@ impl<R, D> Reader<R, D> {
         self.single_frame = true;
     }
 
+    /// Returns a reference to the underlying operation.
+    pub fn operation(&self) -> &D {
+        &self.operation
+    }
+
     /// Returns a mutable reference to the underlying operation.
     pub fn operation_mut(&mut self) -> &mut D {
         &mut self.operation
--- a/src/stream/zio/reader.rs
+++ b/src/stream/zio/reader.rs
@@ -17,6 +17,9 @@ pub struct Reader<R, D> {
 
     single_frame: bool,
     finished_frame: bool,
+
+    peeking: bool,
+    peeked_data: [u8; 4],
 }
 
 enum State {
@@ -39,6 +42,8 @@ impl<R, D> Reader<R, D> {
             state: State::Reading,
             single_frame: false,
             finished_frame: false,
+            peeking: false,
+            peeked_data: [0; 4],
         }
     }
 
@@ -81,7 +86,37 @@ impl<R, D> Reader<R, D> {
     {
         self.operation.flush(&mut OutBuffer::around(output))
     }
+
+    /// Read some data, but do not consume it.
+    pub fn peek_4bytes(&mut self) -> io::Result<[u8; 4]>
+    where
+        R: BufRead,
+        D: Operation,
+    {
+        if !self.peeking {
+            self.reader.read_exact(&mut self.peeked_data)?;
+            self.peeking = true;
+        }
+
+        Ok(self.peeked_data)
+    }
+
+    /// Clear the peeked data.
+    pub fn clear_peeked_data(&mut self) {
+        self.peeking = false;
+    }
+
+    /// Check if there is currently any peeked data.
+    pub fn peeking(&self) -> bool {
+        self.peeking
+    }
+
+    /// Get the peeked data.
+    pub fn peeked_data(&self) -> [u8; 4] {
+        self.peeked_data
+    }
 }
+
 // Read and retry on Interrupted errors.
 fn fill_buf<R>(reader: &mut R) -> io::Result<&[u8]>
 where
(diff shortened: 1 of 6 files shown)


gyscos (commented):
Hi, and thanks for the PR!

Let's start with zstd-safe. I think it would be best to make these free functions, since they have nothing to do with contexts (contexts are not the only way to encode/decode zstd data).

Otherwise, with a few minor details, it looks nice!

gyscos:
On the higher-level `zstd-rs` side, it seems that there's 2 main things:
* Some wrappers for the zstd-safe functions.
* Some re-implementation of read/write skippable frames in a slightly more "streamable" way - no need to buffer the entire frame twice.

The second point in particular is where most changes are required to re-implement quite a bit of logic from the zstd library. I feel like there's quite a lot of added code and complexity, maybe without being worth the added convenience.

In particular, regarding the re-implementation of `read_skippable_frame`:
* Seeking back when an error occurs is a big source of copied code, either from the standard library or from the C zstd library.
* The implementation can stream the `Read` source, but still buffers the entire result.

I would be more comfortable with a solution that:
* Does not seek back on error. If an error happens, the stream is left in an unspecified state. Should simplify the implementation.
* Does not buffer the output, but either provide a `Read` that will only read the skippable content (using `.take(content_size)` on the underlying reader) or stream the frame content to a `Write` given as parameter. This way, users _could_ buffer it if needed.

antoyo:
> * Does not seek back on error. If an error happens, the stream is left in an unspecified state. Should simplify the implementation.

If you do not seek back, how can you detect that there is a skippable frame?
`is_skippable_frame()` requires a `&[u8]`, so it's not streaming.

And an attempt to call `read_skippable_frame()` and checking for an error could leave the stream at a wrong position without being able to continue reading normal frames.

gyscos:
`is_skippable_frame` only needs the `&[u8]` to cover the header (in practice the first 8 bytes), so it's still mostly streaming.

And if something happens in `read_skippable_frame`, then _something_ is wrong, and there's no guarantee the stream is valid anyway.

antoyo:
> `is_skippable_frame` only needs the `&[u8]` to cover the header (in practice the first 8 bytes), so it's still mostly streaming.

But how would you combine this function with reading frames in a streaming fashion without seeking back?

If you read the first 8 bytes (which could be too much bytes IIRC, e.g. I think you can have a frame of 5 bytes) to check if it's a skippable frame, how would you then read the frame (whether it is skippable or not)?

gyscos:
We could rely on the `BufRead` own buffer to "peek" at the first 4 bytes. Though it's true there's no guarantee the `BufRead` will actually fill more than 1 byte. If we do want to support that, one alternative would be to store up to 3 bytes of buffered data in a new `zstd::stream::zio::reader::State::Peeked`, which would be read from before moving back to `State::Reading` - I suspect it might be simpler than the current situation.

Maybe we could go a different way and register a `skippable_frame_handler: Option<Box<dyn FnMut(u32, &mut dyn Reader) -> io::Result<()>>>` callback, which will be called in the regular operations if a skippable frame is found.

antoyo-light:
Ok, so I just pushed with all the changes you asked.

cgbur:
Closes #13 if merged

antoyo:
Ping.

gyscos:
Hi - sorry, have had very little time to look more into this.

antoyo:
Hi.
Any update?

ariel-miculas:
I'm also interested in this feature.

zstd-safe/src/lib.rs
```diff
@@ -804,6 +804,45 @@ impl Default for DCtx<'_> {
     }
 }
 
+#[cfg(feature = "experimental")]
+pub fn read_skippable_frame<C: WriteBuf + ?Sized>(dst: &mut C, magic_variant: &mut u32, input: &[u8]) -> SafeResult {
+    let input_len = input.len();
+    unsafe {
```
ariel-miculas, line 810: I suggest adding SAFETY comments for the unsafe blocks and wrapping the minimum amount of code in the `unsafe` block. See https://rust-lang.github.io/rust-clippy/master/#/undocumented_unsafe_blocks

src/stream/zio/reader.rs
```diff
@@ -113,12 +153,17 @@ where
         loop {
             match self.state {
                 State::Reading => {
+                    let is_peeking = self.peeking;
```
ariel-miculas, line 156: I think you should call the `peeking` function here.

src/stream/zio/reader.rs
```diff
@@ -113,12 +153,17 @@ where
...
+
                     let (bytes_read, bytes_written) = {
                         // Start with a fresh pool of un-processed data.
                         // This is the only line that can return an interruption error.
                         let input = if first {
                             // eprintln!("First run, no input coming.");
                             b""
+                        } else if self.peeking {
```
ariel-miculas, line 164: Either check the `is_peeking` variable or call the `peeking()` function.

ppamorim:
Hi @antoyo, what's the current status of this work? <3 Thank you.

antoyo:
Hi @ppamorim.
I'm not sure since it's been a long time: it seems I might not have addressed the last review.
In any case, I'm not interested in continuing this work, so if anybody is interested in moving this forward, feel free to do so.

ariel-miculas:
Btw there's [zeekstd](https://github.com/rorosen/zeekstd), a Rust implementation of the Zstandard Seekable Format.

ppamorim:
> Btw there's [zeekstd](https://github.com/rorosen/zeekstd), a Rust implementation of the Zstandard Seekable Format.

This crate is nice, but I remember at the time having trouble to use it in my project, I really don't remember why. Need to revisit it.

PR #188, opened 2023-01-24
Optimize zstd-sys vendered build

NobodyXu:
Rm unused function/data, merge constants and use icf=safe to safely
merge identical functions.


gyscos:
Thanks for the work!

PR #193, opened 2023-01-26
refactor `*.c` files discovery in `compile_zstd`

NobodyXu:
Use iterator-style loop instead of for-loop while also reducing size of the `entries: Vec<_>` by filtering out unwanted `entry` before `collect`ing into a `Vec<_>`.

Also, use `Build::files` instead of `Build::file` to avoid the for-loop.

### DIFF
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -101,22 +101,21 @@ fn compile_zstd() {
         let mut entries: Vec<_> = fs::read_dir(dir)
             .unwrap()
             .map(|r| r.unwrap().path())
-            .collect();
-        entries.sort();
-        for path in entries {
+            .filter(|path| path.extension() == Some(OsStr::new("c")))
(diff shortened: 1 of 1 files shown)


zstd-safe/zstd-sys/build.rs
```diff
@@ -101,22 +101,21 @@ fn compile_zstd() {
...
-            if path.extension() == Some(OsStr::new("c")) {
-                config.file(path);
-            }
-        }
+            .filter(|path| {
+               !path
+                    .file_name()
+                    // If path.extension() is Some, then file_name cannot be None.
+                    .unwrap()
+                    .to_string_lossy()
```
gyscos, line 112: We could also use `as_bytes()` to get the backing `&[u8]` and then look for `b"xxhash"` in there, so we avoid allocation even when it's not valid utf-8.

gyscos: Nevermind i thought there was such a contains function.

PR #194, opened 2023-01-26
Fix `generate_bindings` in `build.rs` of zstd-sys

NobodyXu:


gyscos:
Thanks for the PR!

PR #195, opened 2023-01-26
Refactor and optimize `cargo_print`

NobodyXu:
- accept any `&dyn fmt::Display`
 - avoid dup `cargo:` prefix in `content`
 - avoid error caused by `cfg`

### DIFF
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -1,6 +1,6 @@
 use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
-use std::{env, fs};
+use std::{env, fmt, fs};
 
 #[cfg(feature = "bindgen")]
 fn generate_bindings(defs: Vec<&str>, headerpaths: Vec<PathBuf>) {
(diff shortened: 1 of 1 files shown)


zstd-safe/zstd-sys/build.rs
```diff
@@ -215,26 +215,27 @@ fn compile_zstd() {
...
+    cargo_print(&format_args!("root={}", dst.display()));
 }
 
 /// Print a line for cargo.
 ///
 /// If non-cargo is set, do not print anything.
-fn cargo_print(content: &str) {
-    #[cfg(not(feature = "non-cargo"))]
-    println!("{content}");
+fn cargo_print(content: &dyn fmt::Display) {
```
gyscos, line 224: Could we even just take `content: impl fmt::Display` here? So we wouldn't need to add references to each usage.

gyscos: I suppose it could increase the build script's own build time though.

gyscos: In any case something we can look at later.

PR #196, opened 2023-01-26
Fix `main` in `zstd-sys` `build.rs`: Use `var_os` for `CARGO_MANIFEST_DIR`

NobodyXu:


gyscos:
Thanks for the PR!

Issue #190, opened 2023-01-24
Provide ways to disable compression support?


NobodyXu:
Upstream zstd provides build macro `ZSTD_LIB_COMPRESSION` to disable compression support, which will then significantly reduce the build time of the `zstd-sys` crate and reducing the size of the final binary.

However, adding a feature to `zstd-sys` will certainly be a breaking change even if it's enabled by default (could be opt-out via `default-features = false`) and it's likely other crates using this will just switch them on by default.

I'm thinking of a new environment flag `ZSTD_SYS_DISABLE_COMPRESSION`:

When enabled, it will enable `ZSTD_LIB_COMPRESSION` and change all the compression API to just return an error, signaling that is not supported.

But that sounds really hacky and I would rather like other solutions, like using gc-sections in #188 or using cross-languages LTO.

NobodyXu:
Perhaps we should just let the users specify `CFLAGS` to disable compression support #192 ?

nickbabcock:
I'm curious how large the size reduction is, as I'm testing out different compression algorithms in Wasm and as a point of comparison, the default brotli dictionary increases the Wasm size by 3.5x (from 400 KB to 1400 KB). I'm not expecting a similar drastic reduction, but I am curious what the impact would be, as any savings over the current 500 KB zstd wasm build may be greatly appreciated by the web community.

NobodyXu:
@nickbabcock You can use `CFLAG` to do that for now.
Simply define `ZSTD_LIB_COMPRESSION=0` and `ZSTD_LIB_DICTBUILDER=0` as suggested by [modular-build in zstd](https://github.com/facebook/zstd/blob/dev/lib/README.md#modular-build), you should be able to disable compression support, but linking might fail.

Alternatively, you can compile zstd yourself with these flags to try that out.

PR #189 should also reduce the size if you enable feature thin and once `cc` makes a new release containing https://github.com/rust-lang/cc-rs/pull/757 , I will try re-enabling LTO again.

nickbabcock:
Thanks for the tip. I may be a bit slow this morning, but I tried:

```bash
CFLAGS="-DZSTD_LIB_COMPRESSION=0" cargo build -vv --release
```

And I see lines like the following that make me think I'm communicating `ZSTD_LIB_COMPRESSION` correctly.

```
[...]

PR #197, opened 2023-01-27
Fix zstd-sys in thin mode: Replace `ZSTD_LIB_MINIFY` with actual `#define`s

NobodyXu:
since `ZSTD_LIB_MINIFY` is only used in `Makefile` to enable other `#define`s.

Fixed bug introduced in #189 and in pointed out by @nickbabcock in https://github.com/gyscos/zstd-rs/issues/190#issuecomment-1405126770

### DIFF
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -165,14 +165,18 @@ fn compile_zstd() {
     #[cfg(feature = "thin")]
     {
         // Here we try to build a lib as thin/small as possible.
+        // We cannot use ZSTD_LIB_MINIFY since it is only
+        // used in Makefile to define other options.
 
-        // ZSTD_LIB_MINIFY implies:
-        //  - HUF_FORCE_DECOMPRESS_X1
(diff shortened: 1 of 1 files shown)


zstd-safe/zstd-sys/build.rs
```diff
@@ -165,14 +165,18 @@ fn compile_zstd() {
...
+
+        config
+            .define("HUF_FORCE_DECOMPRESS_X1", Some("1"))
+            .define("HUF_FORCE_DECOMPRESS_X2", Some("0"))
+            .define("ZSTD_FORCE_DECOMPRESS_SEQUENCES_SHORT", Some("1"))
+            .define("ZSTD_FORCE_DECOMPRESS_SEQUENCES_LONG", Some("0"))
+            .define("ZSTD_NO_INLINE", Some("1"))
+            // removes the error messages that are
+            // otherwise returned by ZSTD_getErrorName
+            .define("ZSTD_STRIP_ERROR_STRINGS", Some("1"));
```
NobodyXu, line 179: These lines works in a similar fashion to https://github.com/facebook/zstd/blob/f302ad8811643c428c4e3498e28f53a0578020d3/lib/libzstd.mk#L32-L37

PR #198, opened 2023-01-27
Re-enable lto support guarded behind feature flags

NobodyXu:
- fat-lto for fat lto, will override thin-lto if specified
 - thin-lto for thin lto, will fallback to fat-lto is not supported

### DIFF
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -151,16 +151,14 @@ fn compile_zstd() {
         .flag_if_supported("-Wl,--gc-sections")
         .flag_if_supported("-Wl,--icf=safe");
 
-    // TODO: re-enable thin lto when a more robust solution is found.
-    /*
-    if config.get_compiler().is_like_gnu() {
-        config.flag_if_supported("-fwhopr");
-    } else {
-        // gcc has a -flto but not -flto=thin
-        // Apparently this is causing crashes on windows-gnu?
-        config.flag_if_supported("-flto=thin");
+    if cfg!(feature = "fat-lto") {
+        config.flag_if_supported("-flto");
+    } else if cfg!(feature = "thin-lto") {
+        flag_if_supported_with_fallbacks(
+            &mut config,
+            &["-flto=thin", "-flto"],
+        );
     }
-    */
 
     #[cfg(feature = "thin")]
     {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,9 @@ no_asm = ["zstd-safe/no_asm"]
 doc-cfg = []
 zdict_builder = ["zstd-safe/zdict_builder"]
 
+fat-lto = ["zstd-safe/fat-lto"]
+thin-lto = ["zstd-safe/thin-lto"]
+
 [[example]]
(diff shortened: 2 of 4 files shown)


zstd-safe/zstd-sys/build.rs
```diff
@@ -151,16 +151,14 @@ fn compile_zstd() {
...
-        // gcc has a -flto but not -flto=thin
-        // Apparently this is causing crashes on windows-gnu?
-        config.flag_if_supported("-flto=thin");
+    if cfg!(feature = "fat-lto") {
+        config.flag_if_supported("-flto");
+    } else if cfg!(feature = "thin-lto") {
+        flag_if_supported_with_fallbacks(
+            &mut config,
+            &["-flto=thin", "-flto"],
+        );
```
NobodyXu, line 160: The previous bug with lto should be fixed by https://github.com/rust-lang/cc-rs/pull/757

@gyscos I recommend to upgrade to new `cc` once it lands.

But still, I want to be safe here by putting them behind feature flags so that users can decide whether to enable fat-lto or thin-lto.

gyscos:
Thanks for the work!

gyscos:
Hmmm at this moment running `cargo test --features thin-lto` from `zstd-safe` fails.
(Running it from zstd-sys does not matter because the tests don't actually link anything.)

NobodyXu:
> Hmmm at this moment running `cargo test --features thin-lto` from `zstd-safe` fails for me. (Running it from zstd-sys does not matter because the tests don't actually link anything.)

Hmmm both fat-lto and thin-lto works on my M1 macbook.

Well, at least we can override thin-lto with fat-lto, which should work more reliably on more targets.

gyscos:
It also fails with `--features fat-lto`. (Running on a x86_64 linux machine).

Indeed, running it on a M1 macbook running macOS, the tests pass.

NobodyXu:
@gyscos It should get better with https://github.com/rust-lang/cc-rs/pull/757 once cc releases a new version.

gyscos:
Aaah right I had an explicit override for gcc before.

(Was just trying to understand why it's now working _worse_ for me than before disabling LTO.)

NobodyXu:
> Now it only relies on cc detection, which still hasn't released.

My fix for `cc` is actually really trivial if you look at the PR: It simply checks for the exit status.

I don't know why it didn't check for that but this should improve things a bit, when compiler fails without any stderr output.

gyscos:
(This post may sound trivial for anyone used to LTO, but I was not, so here's my journey.)

Soo I just realized this only works if the C lib is compiled by clang (or another LLVM-based compiler), otherwise the intermediate compiled file will only contain gcc's flavor of IR, which rust cannot use.

So at the moment, both `fat-lto` and `thin-lto` only successfully compile if `CC` is set to `clang`, and if the linker is given to proper option.

In addition, `-Clinker-plugin-lto` is probably recommended to actually run LTO, though it appears to at least compile fine if omitted.

**TLDR**, In my case, I had to run:

```
RUSTFLAGS="-Clinker-plugin-lto -Clink-arg=-fuse-ld=lld" CC=clang cargo test -vv --features thin-lto
```

PR #199, opened 2023-01-27
Optimize `compile_zstd` for-loop in zstd-sys build.rs

NobodyXu:
Prevoiusly, we filter based on `entry.path()`, which includes the full path.

But the filtering rules only care about the filename, hence we use `entry.file_name()` here and group two `filter` into one `filter_map`.

### DIFF
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -100,17 +100,19 @@ fn compile_zstd() {
     ] {
         let mut entries: Vec<_> = fs::read_dir(dir)
             .unwrap()
-            .map(|r| r.unwrap().path())
-            .filter(|path| path.extension() == Some(OsStr::new("c")))
-            // Skip xxhash*.c files: since we are using the "PRIVATE API"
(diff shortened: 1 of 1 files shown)


zstd-safe/zstd-sys/build.rs
```diff
@@ -100,17 +100,19 @@ fn compile_zstd() {
...
-            .filter(|path| {
-               !path
-                    .file_name()
-                    // If path.extension() is Some, then file_name cannot be None.
-                    .unwrap()
-                    .to_string_lossy()
-                    .contains("xxhash")
+            .map(Result::unwrap)
+            .filter_map(|entry| {
+                let filename = entry.file_name();
```
gyscos, line 105: Wish we could avoid allocation, but it looks like `DirEntryExt2` is not stable yet :(

gyscos (approved):

zstd-safe/zstd-sys/build.rs
NobodyXu: Yeah and even with that stablised, it would only work on unix.

PR #202, opened 2023-02-01
Rm unused `-Wl` linker args

NobodyXu:


NobodyXu:
I forgot that `cc` does not run the linker and it is run by `rustc` instead.

Sorry about this kind of mistakes.

PR #203, opened 2023-02-05
Fix feature = "thin": Remove defines to disable x2 and long seq

NobodyXu:
Fix https://github.com/cargo-bins/cargo-binstall/pull/768 and add regression test.


NobodyXu:
@gyscos Can you create a new minor release for this bug fix please?

gyscos:
This was released as part of zstd-sys 2.0.7.

NobodyXu:
Thanks!

PR #220, opened 2023-04-05
Update zstd to v1.5.5

### LINKED ISSUE #213: Update downstream zstd to the latest version
Hello! Firstly, thank you for all your efforts.

Probably you have seen [the issue](https://github.com/facebook/zstd/pull/3517) in zstd. Rather sever thing that could led to data corruption. Just letting you know, I believe it is worth to schedule the update if you are using higher compression levels

NobodyXu:
Fixed #213 


dralley:
@gyscos Can you perform a release?

gyscos:
Zstd-sys-2.0.8 should have been released already with zstd 1.5.5?

dralley:
But zstd-safe has zstd-sys 2.0.7 pinned: https://github.com/gyscos/zstd-rs/blob/main/zstd-safe/Cargo.toml#L19

So most people still get 1.5.4

NobodyXu:
> But zstd-safe has zstd-sys 2.0.7 pinned: https://github.com/gyscos/zstd-rs/blob/main/zstd-safe/Cargo.toml#L19

It does not pin zstd-sys to 2.0.7, it requires zstd-sys to be at least 2.0.7

> So most people still get 1.5.4

They just need to run cargo update for that

dralley:
Apologies, the actual problem is that the suffix of zstd-safe doesn't match up with the version of zstd-sys

```
   Compiling zstd-safe v6.0.5+zstd.1.5.4
...
...
   Compiling zstd-sys v2.0.8+zstd.1.5.5
```

That led to confusion.  You are correct that the actual version of zstd being used is 1.5.5

NobodyXu:
Yes, I also find that confusing.

PR #221, opened 2023-04-05
`panic` should not be called on `Drop::drop`

### LINKED ISSUE #219: Encoder/Decoder should not allow for panicking on drop
From https://github.com/gyscos/zstd-rs/issues/80#issuecomment-1271541760

> \[panicking if not flushed yet on drop\] would be consistent with standard implementations (that I'm aware of, like BufWriter) so if developers use it expecting it to function like other writers their code will be correct. Rarely \( - I assume, if the standard library writers adopted the convention\), if there's an error writing the panic will be obvious.

Bluntly put, no the standard implementations don't. They specifically do the other option mentioned (flush on drop) and **explicitly avoid panicking**

https://doc.rust-lang.org/src/std/io/buffered/bufwriter.rs.html#667-674

```rust
impl<W: Write> Drop for BufWriter<W> {
    fn drop(&mut self) {
        if !self.panicked {
            // dtors should not panic, so we ignore a failed flush
            let _r = self.flush_buf();
        }
    }
}
```

I noticed panic on drop being used in this crate because another library I was looking at adopted the same technique based on this crate

From the [`Drop::drop` docs](https://doc.rust-lang.org/std/ops/trait.Drop.html#panics)

> Given that a [`panic!`](https://doc.rust-lang.org/core/macro.panic.html) will call `drop` as it unwinds, any [`panic!`](https://doc.rust-lang.org/core/macro.panic.html) in a `drop` implementation will likely abort.

Here's a demo of said abort

```rust
use std::io::{self, Write};

struct FailOnWrite;

impl Write for FailOnWrite {
    fn write(&mut self, _buf: &[u8]) -> io::Result<usize> {
        Err(io::Error::new(io::ErrorKind::Other, "Can't write"))
    }

    fn flush(&mut self) -> io::Result<()> {
        Err(io::Error::new(io::ErrorKind::Other, "Can't flush"))
    }
}

fn main() {
    let mut enc = zstd::Encoder::new(FailOnWrite, 0).unwrap().auto_finish();
    enc.write_all(b"Now there's buffered data").unwrap();

    panic!("Would be a shame if something panicked while we're unwinding");
}
```

```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
     Running `target/debug/panic-in-drop-bad`
thread 'main' panicked at 'Would be a shame if something panicked while we're unwinding', src/main.rs:19:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "Can't write" }', /home/wintermute/.data/cargo/registry/src/index.crates.io-6f17d22bba15001f/zstd-0.12.3+zstd.1.5.2/src/stream/write/mod.rs:219:20
stack backtrace:
   0:     0x5620cc7bf44a - std::backtrace_rs::backtrace::libunwind::trace::ha271a8a7e1f3d4ef
    ... lots of elided frames
  38:     0x5620cc6514c5 - _start
  39:                0x0 - <unknown>
thread panicked while panicking. aborting.
[1]    28842 IOT instruction (core dumped)  cargo run
```

As you can see the program aborts when it hits the panic while panicking. This likely hasn't been an issue in the wild because of the rarity that the above conditions would all come together naturally

---

Sorry for the wall of text. There are already wrappers in place to handle flushing/finishing on drop, so I'm happy to fix this by ignoring the failures and updating the docs if you agree

CosmicHorrorDev:
Resolves #219

### DIFF
--- a/src/stream/write/mod.rs
+++ b/src/stream/write/mod.rs
@@ -360,7 +360,10 @@ impl<'a, W: Write> Decoder<'a, W> {
 
     /// Returns a wrapper around `self` that will flush the stream on drop.
     pub fn auto_flush(self) -> AutoFlushDecoder<'a, W> {
-        self.on_flush(Box::new(|_result| {}))
+        AutoFlushDecoder {
+            decoder: Some(self),
+            on_flush: None,
+        }
     }
 
     /// Returns a decoder that will flush the stream on drop.
--- a/src/stream/write/mod.rs
+++ b/src/stream/write/mod.rs
@@ -210,19 +210,14 @@ impl<'a, W: Write> Encoder<'a, W> {
     }
 
     /// Returns a wrapper around `self` that will finish the stream on drop.
-    ///
-    /// # Panic
-    ///
-    /// Panics on drop if an error happens when finishing the stream.
     pub fn auto_finish(self) -> AutoFinishEncoder<'a, W> {
-        self.on_finish(Box::new(|result| {
-            result.unwrap();
-        }))
+        self.on_finish(Box::new(|_result| {}))
     }
 
     /// Returns an encoder that will finish the stream on drop.
     ///
-    /// Calls the given callback with the result from `finish()`.
+    /// Calls the given callback with the result from `finish()`. This runs during drop so it's
+    /// important that the provided callback doesn't panic.
     pub fn on_finish<F: FnMut(io::Result<W>)>(
         self,
         f: F,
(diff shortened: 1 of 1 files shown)


src/stream/write/mod.rs
```diff
@@ -210,19 +210,14 @@ impl<'a, W: Write> Encoder<'a, W> {
...
     /// Returns a wrapper around `self` that will finish the stream on drop.
-    ///
-    /// # Panic
-    ///
-    /// Panics on drop if an error happens when finishing the stream.
     pub fn auto_finish(self) -> AutoFinishEncoder<'a, W> {
-        self.on_finish(Box::new(|result| {
-            result.unwrap();
-        }))
+        self.on_finish(Box::new(|_result| {}))
```
NobodyXu, line 214: I think you can avoid that `Box::new` even if it probably allocates a zero-size object:

```suggestion
        AutoFinishEncoder {
            encoder: Some(self),
            on_finish: None,
        }
```

src/stream/write/mod.rs
```diff
@@ -361,19 +356,14 @@ impl<'a, W: Write> Decoder<'a, W> {
...
     /// Returns a wrapper around `self` that will flush the stream on drop.
-    ///
-    /// # Panic
-    ///
-    /// Panics on drop if an error happens when flushing the stream.
     pub fn auto_flush(self) -> AutoFlushDecoder<'a, W> {
-        self.on_flush(Box::new(|result| {
-            result.unwrap();
-        }))
+        self.on_flush(Box::new(|_result| {}))
```
NobodyXu, line 360:
```suggestion
        AutoFlushDecoder {
            decoder: Some(self),
            on_flush: None,
        }
```

NobodyXu (changes requested):

PR #224, opened 2023-04-11
Add zstd_errors.h to bindgen generator when using experimental feature

### LINKED ISSUE #222: zstd-sys: Link ZSTD_getErrorCode
This is useful utility if one wants to examine error code as zstd, despite what its crappy documentation says, has stable enumeration of errors that you can use.
So it would be good to link it too
It is a bit hidden behind in headers https://github.com/facebook/zstd/blob/dev/lib/zstd_errors.h#L106
`zstd.h` doesn't really expose it, but these symbols are public so maybe it would be ok to put `bindgen` on `zstd_errors.h` too?

DoumanAsh:
Fixes #222

Since part of error enum is not stable, I added include only in experimental build
Do you think it is fine to add as part of bindings?


DoumanAsh:
@gyscos I refactored PR to put header under ifdef

PR #226, opened 2023-06-09
Swap libc to `core::ffi`

NiceneNerd:
I do not know if there is any technical reason why this would be a bad idea, but I propose swapping all uses of libc in `zstd-safe` and `zstd-sys` with `core::ffi` in order to support targets which do not fully support libc, such as `aarch64-nintendo-switch-freestanding`.


gyscos:
Hi, and thanks for the PR!

I'm wondering if we still need different bindings between libc (now core:ffi) and std::ffi. Basically, we might be able to remove (or, for backward compatibility, make a no-op) the `std` feature in `zstd-sys`.

gyscos:
Sorry, I just realized this was for zstd-safe, not zstd-sys :upside_down_face:

PR #247, opened 2023-09-23
Allow reusing the same context across multiple Encoders and Decoders

DouglasDwyer:
From profiling, I have found that creating a `zstd::stream::Encoder` and using it to encode a small amount of bytes (less than 100) takes over 200 microseconds per encode on my machine. The vast majority of this time is related to the creation of new zstd contexts, so this PR adds a simple method on `Encoder` and `Decoder` that allows one to supply an external context from `zstd_safe` instead. With these changes, performance is improved - the previously-mentioned benchmark takes less than 2 microseconds per encode.

This PR also makes some minor changes to the WASM shim, as compilation appears to be broken for it.

### DIFF
--- a/src/stream/raw.rs
+++ b/src/stream/raw.rs
@@ -5,6 +5,8 @@
 //!
 //! They are mostly thin wrappers around `zstd_safe::{DCtx, CCtx}`.
 use std::io;
+use std::marker::PhantomData;
+use std::ops::{Deref, DerefMut};
 
 pub use zstd_safe::{CParameter, DParameter, InBuffer, OutBuffer, WriteBuf};
 
@@ -132,7 +134,7 @@ pub struct Status {
 
 /// An in-memory decoder for streams of data.
 pub struct Decoder<'a> {
-    context: zstd_safe::DCtx<'a>,
+    context: MaybeOwned<'a, zstd_safe::DCtx<'a>>,
 }
 
 impl Decoder<'static> {
@@ -148,11 +150,16 @@ impl Decoder<'static> {
         context
             .load_dictionary(dictionary)
             .map_err(map_error_code)?;
-        Ok(Decoder { context })
+        Ok(Decoder { context: MaybeOwned::owned(context) })
     }
 }
 
 impl<'a> Decoder<'a> {
+    /// Creates a new decoder which employs the provided context for deserialization.
+    pub fn with_context<'b: 'a>(context: &'a mut zstd_safe::DCtx<'b>) -> Self {
+        Self { context: MaybeOwned::borrowed(context) }
+    }
+
     /// Creates a new decoder, using an existing `DecoderDictionary`.
     pub fn with_prepared_dictionary<'b>(
         dictionary: &DecoderDictionary<'b>,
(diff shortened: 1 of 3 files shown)


esemeniuc:
would love to see this!

esemeniuc:
bump @gyscos for visibility

gyscos:
Thanks for the PR!

A few questions:
* Since this is all internal API, why separate `MaybeOwned` and `MaybeOwnedInner` types?
* Why store a `* mut ()` + `PhantomData` rather than directly a `&'a mut T`?

DouglasDwyer:
> Since this is all internal API, why separate MaybeOwned and MaybeOwnedInner types?

My instinct was to separate them for readability. However, since it's an internal API, they don't necessarily have to be separate. I will go ahead and change this.

> Why store a * mut () + PhantomData rather than directly a &'a mut T?

The goal for doing this was to ensure that `Encoder<'a>` and `Decoder<'a>` were covariant over `'a`. If `&'a mut T` was used, the function `with_context<'b: 'a>(reader: R, context: &'a mut zstd_safe::DCtx<'b>)` would not compile - it would be impossible to convert a `&'a mut zstd_safe::DCtx<'b>` to a `&'a mut zstd_safe::DCtx<'a>`.

**However!** Looking at this again, `MaybeOwned<T>` actually appears to be unsound (see [this playground for example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=122fcd64e1dd45284683c5bffb2838db)). As such, I am going to refactor `with_context` to just be `fn with_context(reader: R, context: &'a mut zstd_safe::DCtx<'static>)`. I am unable to find a safe way to support having a borrowed `DCtx<'a>` for an arbitrary `'a` - my only idea is to add another lifetime parameter to `Encoder` and `Decoder`, but that breaking change doesn't seem worth it. If someone knows of a better approach, please let me know.

> I don't seem to see these changes anymore, were they already merged through another PR?

Yes, the WASM shim now works properly on `main`.

---

Taking another look at this PR, I feel like the cleanest solution might be to create an entirely separate `Encoder` and `Decoder` class for borrowed contexts, rather than trying to support both owned/borrowed contexts with one struct. But I don't have the time do any more API design than this. Please let me know if you have any other suggestions for pushing the PR forward. Thanks!

DouglasDwyer:
Ping @gyscos

esemeniuc:
I found a workaround that reuses the dictionary: https://gist.github.com/esemeniuc/ad889c617cc4297043fc344136be563f
I'd guess its probably 95% as fast

esemeniuc:
Bump @gyscos

gyscos:
Thanks for the work! Let's start with that, we'll see later if/how we could allow both referencing a dict and a context.

gyscos:
CI failures looks weird - maybe you need to rebase?

gyscos:
Ended up testing and merging the PR locally. Thanks for the work!

lordofpipes:
Is it intentional that an implementation of `with_context` was never added for `zstd::stream::write::Decoder` and `zstd::stream::read::Encoder`? I've been using a fork with this implementation: 0de61edcb0ae383e1d6324d218202f9a820d9d9a and it seems to work fine.

Issue #248, opened 2023-10-07
Deallocations within zstd violate GlobalAlloc's safety contract


SFBdragon:
Hi!

(All testing was done on WASM.)

Here's the [documentation](https://doc.rust-lang.org/stable/std/alloc/trait.GlobalAlloc.html#tymethod.dealloc) for GlobalAlloc's `dealloc`. In particular:

> `layout` must be the same layout that was used to allocate that block of memory.

The C code in `zstd-rs` breaks this, try running the following code in a WASM project (I think you could adapt this and see a similar result on desktop platforms, but I haven't confirmed this).

lib.rs
```rust
use wasm_bindgen::prelude::*;

struct BadAllocator;

unsafe impl std::alloc::GlobalAlloc for BadAllocator {
    unsafe fn alloc(&self, layout: std::alloc::Layout) -> *mut u8 {
        let ptr = (std::arch::wasm32::memory_grow::<0>(layout.size() / 64 / 1024 + 1) * 64 * 1024) as *mut u8;

        log("alloc: ptr, size");
        log_u32(ptr as u32);
        log_u32(layout.size() as u32);

        ptr
    }

    unsafe fn dealloc(&self, ptr: *mut u8, layout: std::alloc::Layout) {
        log("dealloc: ptr, size");
        log_u32(ptr as u32);
        log_u32(layout.size() as u32);
    }
}

#[global_allocator]
static BADALLOC: BadAllocator = BadAllocator;


#[wasm_bindgen]
extern "C" {
    #[wasm_bindgen(js_namespace = console)]
    fn log(s: &str);

    #[wasm_bindgen(js_namespace = console, js_name = log)]
    fn log_u32(a: u32);
}

#[wasm_bindgen]
pub fn compress(data: &[u8], level: i32) -> Vec<u8> {
    zstd::bulk::compress(data, level).unwrap()
}
```

index.js
```js
import * as wasm from "wasm-alloc-bench";

var array = new Array(10000);

for (let i = 0; i < array.length; i++) {
    array[i] = i % 256;
}

wasm.compress(array, 0);
```

Which produces the output:

![console output](https://github.com/gyscos/zstd-rs/assets/46199751/a3ab0ac5-30fe-42a9-bfd6-c52af5056bba)

Notice that there are two deallocations of size 1, which were allocated with an entirely different size (`layout` contains the size).

So why does the code normally work?
[...]

PR #249, opened 2023-10-09
fix wasm_shim allocation behavior as per #248

SFBdragon:


zstd-safe/zstd-sys/src/wasm_shim.rs
```diff
@@ -14,17 +29,33 @@ pub extern "C" fn rust_zstd_wasm_shim_calloc(
     nmemb: usize,
     size: usize,
 ) -> *mut c_void {
+    // the strategy for rust_zstd_wasm_shim_malloc is used, described there
```
gyscos, line 32: Should we maybe just call `rust_zstd_wasm_shim_malloc(nmemb * size)`?

SFBdragon: Sounds good, although this prompted me to check the c docs: turns out calloc expects its memory to be zeroed.

gyscos:
> Sounds good, although this prompted me to check the c docs: turns out calloc expects its memory to be zeroed.

Woops indeed! Guess on wasm there's no special CoW zero page we could use, so let's just zero things out manually here?

SFBdragon:
I hope you can squash my messy commit history :)

(Everything is still using std for now)

Issue #253, opened 2023-11-22
Incorrect minimum supported rust version


hoxxep:
Using [`cargo msrv`](https://github.com/foresterre/cargo-msrv) suggests that the actual minimum supported rust version is 1.66.1 due to the following compile failure in the jobserver v0.1.27 sub-dependency required by the `cc/parallel` feature.

```
Check for toolchain '1.64.0-aarch64-apple-darwin' failed with:
┌──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ Compiling libc v0.2.150                                                                                                                                          │
│    Compiling pkg-config v0.3.27                                                                                                                                  │
│    Compiling zstd-safe v7.0.0 (/Users/liam/RustroverProjects/zstd-rs/zstd-safe)                                                                                  │
│    Compiling jobserver v0.1.27                                                                                                                                   │
│ error[E0412]: cannot find type `RawFd` in `std::os::fd`                                                                                                          │
│  --> /Users/liam/.cargo/registry/src/github.com-1ecc6299db9ec823/jobserver-0.1.27/src/error.rs:2:27                                                              │
│   |                                                                                                                                                              │
│ 2 | type RawFd = std::os::fd::RawFd;                                                                                                                             │
│   |                           ^^^^^ not found in `std::os::fd`                                                                                                   │
│   |                                                                                                                                                              │
│ help: consider importing this type alias                                                                                                                         │
│   |                                                                                                                                                              │
│ 1 | use std::os::unix::prelude::RawFd;                                                                                                                           │
│   |                                                                                                                                                              │
│ help: if you import `RawFd`, refer to it directly                                                                                                                │
│   |                                                                                                                                                              │
│ 2 - type RawFd = std::os::fd::RawFd;                                                                                                                             │
│ 2 + type RawFd = RawFd;                                                                                                                                          │
│   |                                                                                                                                                              │
│                                                                                                                                                                  │
│ error[E0603]: module `fd` is private                                                                                                                             │
│  --> /Users/liam/.cargo/registry/src/github.com-1ecc6299db9ec823/jobserver-0.1.27/src/error.rs:2:23                                                              │
│   |                                                                                                                                                              │
│ 2 | type RawFd = std::os::fd::RawFd;                                                                                                                             │
│   |                       ^^ private module                                                                                                                      │
│   |                                                                                                                                                              │
│ note: the module `fd` is defined here                                                                                                                            │
[...]

PR #254, opened 2023-11-22
Fix minimum supported rust version to 1.66

hoxxep:
Matching PR for issue https://github.com/gyscos/zstd-rs/issues/253 bumping the minimum supported rust version to 1.66 in Cargo.toml.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ version = "0.13.0"
 exclude = ["assets/*.zst", "/.github"]
 readme = "Readme.md"
 edition = "2018"
-rust-version = "1.66.1"
+rust-version = "1.66"
 
 [package.metadata.docs.rs]
 features = ["experimental", "zstdmt", "zdict_builder", "doc-cfg"]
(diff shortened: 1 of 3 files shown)


dtolnay (commented):
I think the original values before this PR are correct. All of these packages build successfully with Rust 1.64 and jobserver 0.1.26. Why forbid folks from building them with Rust 1.64?

A better fix would be to add rust-version metadata to jobserver's Cargo.toml to improve the error message when using a too new jobserver with an old compiler, but I see that's been done already by https://github.com/rust-lang/jobserver-rs/pull/59.

hoxxep:
Good point about it forbidding users from building with 1.64. I was approaching it as being slightly misleading as `cargo update` is unable to resolve a working dependency tree without manually searching for a compatible version of jobserver (and hence, `cargo msrv` sees the MSRV of this crate as 1.66).

I've since had some more discussion on a related jobserver issue (https://github.com/rust-lang/jobserver-rs/pull/62#issuecomment-1823024200) where it seems that MSRV-aware dependency resolution in cargo would be the real solution (https://github.com/rust-lang/cargo/issues/9930).

hoxxep:
Closing this issue and PR, as @dtolnay points out the MSRV of dependencies is linked but mostly independent of the MSRV of this crate.

PR #261, opened 2024-02-03
Bump zstd-sys version in zstd-safe

ryanking13:
Hi, we are building a library targeting wasm32-unknown-emscripten that uses zstd, and noticed that recent zstd-sys updates regarding emscripten (https://github.com/gyscos/zstd-rs/pull/209, https://github.com/gyscos/zstd-rs/commit/92f87dca4987555fe8a87504dea68b37a29d62b9) are not applied to zstd-safe 7.0.0. It would be great to have a new release that includes these updates.


NobodyXu:
It's a patch update, you can simply update your `Cargo.lock`

ryanking13:
Oh, I see. Thanks, that works.

PR #264, opened 2024-02-27
Add `vendored` feature to zstd-sys for providing pre-built zstd libs

abrisco:
I would like to be able to provide my own `libzstd.a` when building `zstd-safe`. This change introduces the `vendored` feature in combination with the following environments to support this:

- `ZSTD_SYS_VENDORED`: An extra flag to enable the vendored search even if the feature is not enabled
- `ZSTD_SYS_VENDORED_LIBS`: A space delimited string of paths to compiled zstd libraries
- `ZSTD_SYS_VENDORED_INCLUDE`: The path to the zstd include directory where headers like `zstd.h` can be found.

### DIFF
--- a/zstd-safe/zstd-sys/Cargo.toml
+++ b/zstd-safe/zstd-sys/Cargo.toml
@@ -79,6 +79,7 @@ no_asm = [] # Disable ASM files (only on amd64 for decompression)
 zdict_builder = [] # Enable dictionary building (dictionary _using_ is always supported).
 no_wasm_shim = [] # Disable wasm shims (in case your wasm toolchain includes a C stdlib).
 seekable = [] # Enable support of the seekable format
+vendored = [] # Allow users to provide explicit lib and inlcude paths for the zstd C library.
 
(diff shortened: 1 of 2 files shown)


abrisco:
The naming of the feature was inspired by https://docs.rs/openssl/latest/openssl/#vendored

gyscos:
Hi, and thanks for the report!

Is there any way this could be done with the current pkg_config feature?

abrisco:
> Hi, and thanks for the report!
> 
> Is there any way this could be done with the current pkg_config feature?

Hey! I don't know how I missed your message (and have been surviving on a patch for a while) but `pkg_config` requires... well... `pkg-config` (the binary) to be available. I'm using this library with Bazel which provides a separate mechanism for providing the exact files or directories for linking and includes that I want to be able to wire into the build script.

jhult:
Related: #309

gyscos:
Thanks! The vendored feature should now be available.

PR #274, opened 2024-04-12
Fix wasm32-unknown-unknown support

### LINKED ISSUE #271: zstd-sys for wasm32-unknown-unknown is broken
`zstd-sys` version is 2.0.10 and error.

```bash
Error: Compiling your crate to WebAssembly failed
Caused by: Compiling your crate to WebAssembly failed
Caused by: failed to execute `cargo build`: exited with exit code: 101
  full command: "cargo" "build" "--lib" "--release" "--target" "wasm32-unknown-unknown"
```

I set
```toml
zstd-sys = "=2.0.9"
```
and build successed.

jbms:
`XXH_STATIC_ASSERT` appears to already have a reasonable definition, and defining it to `0` breaks uses of it.

I'm not sure why that definition was added originally.

This also fixes the github action workflow to actually build the wasm target; previously it was just building the native target which allowed this issue to slip through.

Fixes #271.

### DIFF
--- a/.github/workflows/wasm.yml
+++ b/.github/workflows/wasm.yml
@@ -22,6 +22,6 @@ jobs:
         submodules: recursive
 
     - name: Build
-      run: cargo build --verbose
+      run: cargo build --verbose --target wasm32-unknown-unknown
     - name: Build with feature thin
-      run: cargo build --verbose --features thin
+      run: cargo build --verbose --features thin --target wasm32-unknown-unknown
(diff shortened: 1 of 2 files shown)


xangelix:
@gyscos apologies for reaching out directly, but could you take a look at this? should be a quick fix for wasm32 along with https://github.com/gyscos/zstd-rs/pull/275

juntyr:
I, unfortunately, have to bump this issue as well as pinning zstd-sys to `=2.0.9` is no longer possible for me as now other dependencies, which don't explicitly target wasm, have now upgraded their zstd-sys dependencies, resulting in conflicts with my pin. @gyscos, once you find some time, it would be amazing if you could accept this minimal PR and publish a fix release. Thank you for all your work on this crate!

juntyr:
Thank you so much! Would you be able to publish a patch release for  zstd-sys to crates.io / what’s your rough timeline until the next release?

Issue #282, opened 2024-06-02
Please provide safe bindings to the thread pool mechanism


joshtriplett:
zstd supports sharing thread pools between threads. You can call `ZSTD_createThreadPool` to create a pool of a given size, `ZSTD_CCtx_refThreadPool` to assign a thread pool to a `CCtx`, and `ZSTD_freeThreadPool` to free a thread pool when no longer in use.

Please consider providing safe bindings to this mechanism. One option would be to handle them like dictionaries, and require that the thread pool have a lifetime outliving the `CCtx`.

PR #283, opened 2024-06-02
Support thread pools

joshtriplett:
Create a safe `ThreadPool` type wrapping `ZSTD_threadPool`. Allow binding a `&ThreadPool` to a `CCtx`, which requires that the lifetime of the `&ThreadPool` outlive the `CCtx`.

Fixes: #282


gyscos:
Hi, and thanks for the work! This looks perfect!

PR #289, opened 2024-07-08
Update `need_wasm_shim` logic for `wasm32-wasip1`

Mrmaxmeier:
Hi,

the `wasm32-wasi` target is being renamed to `wasm32-wasip1` and the
`wasm32-wasi` target will be removed in the future. It looks like the
shim will also be useful (required?) for `wasm32-wasip2`, so I've included all future `wasm32-wasi*` targets for now.

Thanks!

### DIFF
--- a/src/stream/functions.rs
+++ b/src/stream/functions.rs
@@ -55,5 +55,5 @@ where
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


gyscos:
Thanks for the work!

PR #294, opened 2024-08-02
release: zstd 0.13.3, zstd-safe 7.3.0, zstd-sys 2.0.13

shekhirin:
Includes the fix for compilation on new Wasm targets https://github.com/gyscos/zstd-rs/commit/fc3a73100b6d3f2ff6dd3da0ff14dd9185ba02f0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ categories = ["compression", "api-bindings"]
 license = "MIT"
 name = "zstd"
 repository = "https://github.com/gyscos/zstd-rs"
-version = "0.13.2"
+version = "0.13.3"
 exclude = ["assets/*.zst", "/.github"]
 readme = "Readme.md"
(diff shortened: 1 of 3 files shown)


gyscos:
Thanks for the PR!

It seems only zstd-safe had a change in the wasm detection (not even any API change). I released zstd-safe 7.2.1, I don't think we need to release the other two.

shekhirin:
@gyscos I see, thanks for the release!

It seems like `zstd-sys` also needs a release: updating `zstd-safe` didn't seem to work, and the change of `build.rs` was in `zstd-sys`.

gyscos:
Ugh indeed, putting `zstd-sys` inside the `zstd-safe` folder is really confusing me, need to fix it some day.
Thanks for noticing! zstd-sys 2.0.13 out the door.

PR #297, opened 2024-08-17
Don't allow writing to finished Encoder

anacrolix:
If an Encoder is finished, it will swallow writes without error but never let you flush or finish those.


gyscos:
Thanks for the PR!

anacrolix:
Thanks for the crate, mate

Issue #300, opened 2024-08-30
Make the buffer size configurable?


ksolana:
The writer::new function has hardcoded buffer value.

```rust
    /// Creates a new `Writer`.
    ///
    /// All output from the given operation will be forwarded to `writer`.
    pub fn new(writer: W, operation: D) -> Self {
        Writer {
            writer,
            operation,

            offset: 0,
            // 32KB buffer? That's what flate2 uses
            buffer: Vec::with_capacity(32 * 1024),

            finished: false,
            finished_frame: false,
        }
    }
```

Can we make the buffer size configurable? I'd like to experiment with different buffer sizes and see if that helps with compression ratio/speed.

gyscos:
Hi, and thanks for the request!

Indeed, it would likely make sense to expose a way to set a custom buffer size. I'm surprised by the 12% saving reported in that hail ticket though.

PR #307, opened 2024-10-08
Make the buffer size configurable

ksolana:
Fixes: #300


gyscos:
Hi, and thanks for the work!

This looks reasonable, but you could already call `with_output_buffer`, right? This is just convenience to not create the `Vec` yourself?

gyscos:
Sorry for the delay, and thanks for the work!

PR #308, opened 2024-10-21
OSS-Fuzz: Add fuzzer for OSS-Fuzz integration

arthurscchan:
This PR creates a `cargo fuzz` directory along with a fuzzing harness. The aim is to include the `zstd-safe` module for fuzzing under OSS-Fuzz.

Fuzzing is essentially a stress testing technique used to uncover bugs in software, and OSS-Fuzz is a free service run by Google that provides continuous fuzzing for important open-source projects. Integrating `zstd-safe` would be beneficial in identifying potential memory corruption issues.

The only requirement for integration at this stage is an email linked to a Google account, which will be used to receive notifications when bugs are detected.

An initial integration has been submitted via https://github.com/google/oss-fuzz/pull/12626, which relies on the fuzzer being hosted upstream. This is the primary objective of this PR.

### DIFF
--- a/zstd-safe/fuzz/fuzz_targets/zstd_fuzzer.rs
+++ b/zstd-safe/fuzz/fuzz_targets/zstd_fuzzer.rs
@@ -26,7 +26,7 @@ fuzz_target!(|data: &[u8]| {
     for level in 0..=20 {
         if let Ok(written) = zstd_safe::compress(&mut buffer[..], data, level) {
             let compressed = &buffer[..written];
-            let mut decompressed = vec![0u8; 2048];
+            let mut decompressed = vec![0u8; buffer_size];
             let _ = zstd_safe::decompress(&mut decompressed[..], compressed).unwrap_or_else(|_| 0);
         }
     }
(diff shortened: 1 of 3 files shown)


zstd-safe/fuzz/fuzz_targets/zstd_fuzzer.rs
```diff
@@ -0,0 +1,87 @@
...
+    let mut cctx_stream = zstd_safe::CCtx::default();
+    let mut dctx_stream = zstd_safe::DCtx::default();
+    let mut in_buffer = zstd_safe::InBuffer::around(data);
+    let mut out_buffer = zstd_safe::OutBuffer::around(&mut buffer[..]);
+
+    if let Ok(_) = cctx_stream.compress_stream(&mut out_buffer, &mut in_buffer) {
+        let mut decompressed_stream = vec![0u8; 2048];
+        let mut out_buffer_stream = zstd_safe::OutBuffer::around(&mut decompressed_stream[..]);
+        let mut in_buffer_stream = zstd_safe::InBuffer::around(out_buffer.as_slice());
+        let _ = dctx_stream.decompress_stream(&mut out_buffer_stream, &mut in_buffer_stream).unwrap_or_else(|_| 0);
```
gyscos, line 64: This will silently ignore any decompression failure, is that intended?

gyscos (approved):

zstd-safe/fuzz/fuzz_targets/zstd_fuzzer.rs
arthurscchan: I am thinking if we provide random bytes for fuzzing, there is a high possibility to result in decompression failure and those failure is kinda expected. Thus it is not the target of fuzzing and so I decided to ignore them.

PR #309, opened 2024-11-11
Add `vendored` cargo feature

kpcyrd:
This is already the default behavior of the `zstd` crate, but this feature allows the following:

```toml
[feature]
vendored = ["openssl/vendored", "zstd/vendored"]

[dependencies]
openssl = "0.10"
zstd = { version = "0.13", features = ["pkg-config"] }
```

With a Cargo.toml like this, the default behavior would by dynamic linking for both openssl and zstd, and a static binary could be built with:

```sh
cargo build --target x86_64-unknown-linux-musl -F vendored
```

At the moment, the Cargo.toml would need to be written like this:

```toml
[feature]
default = ["zstd/pkg-config"]
vendored = ["openssl/vendored"]

[dependencies]
openssl = "0.10"
zstd = "0.13"
```

And built like this:
```sh
cargo build --target x86_64-unknown-linux-musl --no-default-features -F vendored
```

I also made sure that, even with the `vendored` feature set, one could still opt into dynamic linking with:

```sh
ZSTD_SYS_USE_PKG_CONFIG=1 cargo build
```

The options take precedence in this order (lowest to highest):

1. default (vendored)
2. `pkg-config` (dynamically linked)
3. `vendored` (vendored)
4. `ZSTD_SYS_USE_PKG_CONFIG=1` (dynamically linked)

Thanks!


orhun (approved):

palfrey:
I'm hitting a variant of https://users.rust-lang.org/t/undefined-reference-to-memcpy-chk-with-target-x86-64-unknown-linux-musl/113567 and this PR would fix things. What's needed to get this merged?

PR #322, opened 2025-02-16
Add Send and Sync markers for AdvancedSeekable

rorosen:
Right now `AdvancedSeekable` isn't thread-safe as it contains a raw pointer to `F`. This adds impls to make `AdvancedSeekable` `Send` or `Sync`, if `F` is `Send` or `Sync`.


gyscos:
Thanks for the PR!

PR #324, opened 2025-02-20
Update zstd library to v1.5.7

JakubOnderka:


zstd-safe/zstd-sys/src/bindings_zstd_experimental.rs
```diff
@@ -75,6 +75,53 @@ pub const ZSTD_LDM_HASHRATELOG_MIN: u32 = 0;
 pub const ZSTD_TARGETCBLOCKSIZE_MIN: u32 = 1340;
 pub const ZSTD_TARGETCBLOCKSIZE_MAX: u32 = 131072;
 pub const ZSTD_SRCSIZEHINT_MIN: u32 = 0;
+pub const ZSTD_SRCSIZEHINT_MAX: u32 = 2147483647;
```
gyscos, line 78: Interestingly my bindgen cli doesn't generate bindings for this value.

PR #325, opened 2025-02-20
fix: update ZSTD_ParamSwitch_e enum in zstd-safe

tchardin:
#324 is breaking zstd-safe it seems, submitting a fix if that helps. Thanks


gyscos:
Thanks for the PR!

PR #328, opened 2025-02-21
Add cfg attribute for advanced seekable impls and make seek table creation safe

rorosen:
Two minor fixes

- The impls on `AdvancedSeekable` had no `cfg` macros, leading to a compilation failure when the `seekable` feature was enabled but not `std`, e.g.:

```text
cargo t -F seekable
   Compiling zstd-safe v7.2.3 (/home/rob/misc/zstd-rs/zstd-safe)
   Compiling zstd-sys v2.0.14+zstd.1.5.7 (/home/rob/misc/zstd-rs/zstd-safe/zstd-sys)
error[E0412]: cannot find type `AdvancedSeekable` in this scope
   --> src/seekable.rs:458:25
    |
458 | unsafe impl<F> Send for AdvancedSeekable<'_, F> where F: Send {}
    |                         ^^^^^^^^^^^^^^^^ not found in this scope

error[E0412]: cannot find type `AdvancedSeekable` in this scope
   --> src/seekable.rs:459:25
    |
459 | unsafe impl<F> Sync for AdvancedSeekable<'_, F> where F: Sync {}
    |                         ^^^^^^^^^^^^^^^^ not found in this scope

For more information about this error, try `rustc --explain E0412`.
error: could not compile `zstd-safe` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error: could not compile `zstd-safe` (lib test) due to 2 previous errors
```

- With zstd v1.5.7 a segfault that previously could occur in `SeekTable::try_from_seekable` is fixed, hence the function can be made safe now

### DIFF
--- a/zstd-safe/src/seekable.rs
+++ b/zstd-safe/src/seekable.rs
@@ -599,17 +599,11 @@ unsafe impl Send for SeekTable {}
 unsafe impl Sync for SeekTable {}
 
 impl SeekTable {
-    // `ZSTD_seekTable_create_fromSeekable` causes a segmentation fault when called with an
-    // uninitialized Seekable. This function is safe once the issue is resolved and released
-    // upstream.
-    // See https://github.com/facebook/zstd/issues/4200 and https://github.com/facebook/zstd/pull/4201
     /// Try to create a `SeekTable` from a `Seekable`.
     ///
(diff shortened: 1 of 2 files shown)


gyscos:
Thanks for the pr! We'll need to add this feature combination to the CI tests next.

PR #329, opened 2025-02-21
Add zstd-safe to CI tests

rorosen:
Follow-up to #328 

Build and test zstd-safe with different feature combinations in CI.
I only added it to Linux for now since I'm not sure about the others. Let me know if this was meant to be different.


gyscos:
Thanks for the PR!

PR #332, opened 2025-02-24
Release zstd-sys and zstd-safe

rorosen:
Would you mind to release new versions of `zstd-sys` and `zstd-safe`? I would like to use zstd-safe with the seekable feature but this is currently not possible with version `7.2.3` form crates.io (due to https://github.com/gyscos/zstd-rs/pull/330).


gyscos:
Sorry for the delay! Things should now have been published.

PR #335, opened 2025-03-20
enhancement: expose `ZSTD_getFrameProgression` through `CCtx`

tobz:
This PR exposes the ability to query the frame progression state (`ZSTD_getFrameProgression`) through `CCtx`.

I opted to simply return `ZSTD_frameProgression` directly rather than creating a newtype or anything.


tobz:
Ah, I totally missed that! :)

Will update the PR.

PR #338, opened 2025-05-02
Add missing `with_context` constructors

lordofpipes:
It seems like it was intended for these to be added in #247. I've been using this implementation and it seems to work fine.


gyscos:
Thanks for the PR!

PR #346, opened 2025-07-24
Don't force non-additive bindgen runtime feature

JadedBlueEyes:


gyscos:
Thanks! We ended up making bindgen non-default.

PR #355, opened 2025-12-19
Add RISC-V RVV optimization support for zstd C library compilation

Dayuxiaoshui:
This commit adds automatic RISC-V RVV (RISC-V Vector Extension) optimization support when building the zstd C library on RISC-V 64-bit platforms.

Changes:

- Detect riscv64 target architecture in build.rs

- Automatically apply -march=rv64gcv flag (with fallback to -march=rv64gc)

- Add -O3 optimization level for RISC-V builds

- Use flag_if_supported_with_fallbacks to gracefully handle unsupported flags

The optimization is automatically applied when compiling on riscv64 targets, with graceful fallback if the compiler doesn't support RVV extensions.

Tested on:

- RISC-V 64-bit platform (riscv64)

- All tests pass (32 unit tests + 1 integration test + 4 doc tests)

- Build successful with both debug and release profiles

This enables better performance for zstd compression/decompression on RISC-V platforms that support vector extensions.

### DIFF
--- a/zstd-safe/zstd-sys/build.rs
+++ b/zstd-safe/zstd-sys/build.rs
@@ -178,7 +178,10 @@ fn compile_zstd() {
             &mut config,
             &["-march=rv64gcv", "-march=rv64gc"],
         );
-        config.flag_if_supported("-O3");
+        let profile = env::var("PROFILE").unwrap_or_default();
+        if profile == "release" {
+            config.flag_if_supported("-O3");
+        }
     }
 
     if cfg!(feature = "fat-lto") {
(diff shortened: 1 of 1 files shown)


gyscos:
Hi, and thanks for the PR!

Everything looks fine - I'm just wondering if the `-O3` should be conditional on the optimization level requested, rather than always applied?

Dayuxiaoshui:
Thanks for the feedback! You're absolutely right. I've updated the patch to only apply `-O3` for release builds by checking the `PROFILE` environment variable. This respects the user's build configuration - debug builds will use default optimization levels while release builds get the performance benefits of `-O3`. The patch has been updated accordingly.

PR #358, opened 2026-01-22
Update license info in zstd-sys `Cargo.toml` to SPDX format & include BSD-3-Clause

edocli:


gyscos:
Hi, and thanks for the PR!

Let's try to keep the licensing as simple as possible. zstd itself (the C lib) seems to be released as BSD or GPLv2. Let's just use the same license for the bindings then, so we're exactly as permissive. We can drop the MIT/Apache (leftover from the start of the lib) and just keep `BSD-3-Clause or GPL-2.0`.

gyscos:
Replaced by #360

PR #361, opened 2026-03-10
feat: allow building using cmake

### LINKED ISSUE #229: Error cross-compiling from M1 MacOS to Linux
I'm trying to build this crate as part of a project I'm working on. All the other rust crates build fine when cross-compiling (I need to target `aarch64-unknown-linux-musl`), but attempting to use this crate fails at building `zstd-sys` with a very long error message (>1800 lines, [err.log](https://github.com/gyscos/zstd-rs/files/12043743/err.log)), containing many repetitions of `#error Unsupported architecture` and `#error architecture not supported` from various `.c` and `.h` files being built.

This error is specific to cross-compiling, I can build on my M1 Mac targeting itself, or my Linux box targeting itself and those builds work fine. It's only when I attempt to cross-compile that I get issues. I also get the same error building the `train` example in this repo (which is what I did to generate the error I attached above).

Judging by the output I'm seeing, I suspect the compilation process is pulling in headers for MacOS instead of the correct headers for targeting Linux. However, I don't see any options in this crate that control where those headers come from, and I don't have these issues with building any other rust crates when I cross-compile, so I suspect it's specific to the headers being pulled in here. How can I control what headers it includes, to fix this error?

I'm able to build the zstd submodule by just running `make`, but I don't know how to ask `make` to cross-compile its build when running manually (I'm not very experienced with C/C++, so my ability to diagnose issues on that side is limited).

jhult:
- Defaults to `cc` but adds new feature `cmake`
- Adds a new cross-compilation workflow (using https://github.com/rust-cross/cargo-zigbuild) to prove this works (you can see a passing CI run [here](https://github.com/jhult/zstd-rs/actions/runs/22903809803))

This resolves #229.


jhult:
@gyscos - any thoughts on this?

jhult:
@gyscos - have you had a chance to look at this?

gyscos:
Thanks for the work! 

Ended up merging it in https://github.com/gyscos/zstd-rs/commit/c9ecfc73437913f33034f09d94d7b5a3d3680c0e. The zig part was left out so far, might need a separate PR.

PR #367, opened 2026-06-24
Mark unsound safe wasm shims qsort and memcmp as unsafe

Manishearth:
Found during an agent-driven safety review. I assume these functions aren't _actually_ supposed to be callable from Rust, but they currently are.

This is technically a breaking change but I think anyone calling these from Rust is probably doing it wrong, unless I've misunderstood the API.


gyscos:
Indeed, thanks for the PR!

PR #371, opened 2026-08-26
Exclude bindgen from zstd-sys default features

### LINKED ISSUE #370: Exclude bindgen from default features
bindgen is hard to use outside of Cargo (e.g. in Bazel).

Some crates in the workspace (e.g. `rdkafka-sys`) pull zstd with default features, so they enable bindgen feature on the workspace, even it the other crates do not use crates like `rdkafka-sys`.

Those projects can be hunted one by one and patched to exclude bindgen, but easier option would be ship zstd with bindgen off by default.

stepancheg:
Keep the remaining defaults (legacy, zdict_builder). bindgen stays available as an opt-in feature; pre-generated bindings are used otherwise.

bindgen makes hermetic compilation (e.g. Bazel) and cross-compilation hard, because it needs libclang at build time.

Fixes #370

AI assistance was used.

### DIFF
--- a/zstd-safe/zstd-sys/Cargo.toml
+++ b/zstd-safe/zstd-sys/Cargo.toml
@@ -65,7 +65,7 @@ version = "1.0.45"
 features = ["parallel"]
 
 [features]
-default = ["legacy", "zdict_builder", "bindgen"]
+default = ["legacy", "zdict_builder"]
 
 debug = [] # Enable zstd debug logs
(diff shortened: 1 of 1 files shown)


gyscos:
Thanks for the work! The bindgen feature is mostly useful for targets without pre-built headers, but I suppose it's easier to _add_ the feature when needed than to remove it.

